PR #4, opened 2019-04-01
modify certificaterequest to make it cloud-agnostic

joelddiaz:
the changes burry the cloud-specific pieces of the CRD into spec.PlatformSecrets.

this should make it easier to add additional cloud support in the future if needed.

also change the name of the baseDomain parameter to make it clear that it is not the primary certificate name.


joelddiaz:
@csrwng see if this looks like it addresses what we discussed this morning

@tparikh this is partly the result of the conversation we had regarding the baseDomain field last week

csrwng:
@joelddiaz I had suggested `AuthenticationDomain` but what you have also sounds fine.

pkg/apis/certman/v1alpha1/certificaterequest_types.go
```diff
@@ -108,6 +108,19 @@ type CertificateRequestList struct {
...
+type PlatformSecrets struct {
+	// +optional
+	AWS *AWSPlatformSecrets `json:"aws,omitempty"`
+}
+
+// AWSPlatformSecrets contains secrets for clusters on the AWS platform.
+type AWSPlatformSecrets struct {
+	// Credentials refers to a secret that contains the AWS account access
+	// credentials.
+	Credentials corev1.LocalObjectReference `json:"credentials"`
```
tparikh: @joelddiaz Credentials should be of type `corev1.ObjectReference` as the credentials won't be in the same namespace as the operator. I don't think `corev1.LocalObjectReference` allows you to specify `Namespace`.

joelddiaz: i don't believe we plan to allow a CertificateRequest in namespace A to point to a secret in namespace B, so the namespace can be inferred from the namespace on the CertificateRequest object itself. if i'm wrong in my understanding here, please let me know.

tparikh:
lgtm

PR #5, opened 2019-04-03
Clusterdeployment controller

joelddiaz:
controller that watches ClusterDeployments and creates any required CertificateRequests

test cases to cover the initial use-case of having one CertificateBundle that covers the entire ClusterDeployment


joelddiaz:
@csrwng probably worth reviewing that 2nd and 3rd commit in this PR

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -0,0 +1,260 @@
...
+	// get all CRs in the cluster's namespace
+	currentCRs := &certmanv1alpha1.CertificateRequestList{}
+	if err := r.client.List(context.TODO(), &client.ListOptions{Namespace: cd.Namespace}, currentCRs); err != nil {
+		logger.Error(err, "error listing current CertificateRequests")
+		return certReqsForCluster, err
+	}
+
+	// now filter out the ones that are owned by the cluster we're processing
+	for i, cr := range currentCRs.Items {
+		for _, owner := range cr.GetOwnerReferences() {
```
csrwng: You can use metav1.IsControlledBy()

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -0,0 +1,260 @@
...
+			}
+		} else {
+			// update or no updated needed
+			if !reflect.DeepEqual(currentCR.Spec, desiredCR.Spec) {
+				if err := r.client.Update(context.TODO(), currentCR); err != nil {
+					logger.Error(err, "error updating certificaterequest", "certrequest", currentCR.Name)
+					return err
+				}
+			} else {
+				logger.Info("no updated needed for certificaterequest", "certrequest", desiredCR.Name)
```
csrwng: s/updated/update/

csrwng: You also likely meant to have logger.WithField("certrequest", desiredCR.Name).Info(...)

joelddiaz: @csrwng i'm just following along with what the default logger library is for operator-sdk (go-logr)...it takes the key=value pairs without having to do the .WithField(): https://github.com/openshift/certman-operator/blob/master/vendor/github.com/go-logr/logr/logr.go#L111

pkg/controller/clusterdeployment/clusterdeployment_controller.go
joelddiaz: done. (also updated the mocking in the test cases to that metav1.IsControlledBy() works)

pkg/controller/clusterdeployment/clusterdeployment_controller.go
csrwng: sorry @joelddiaz I simply assumed you were using the sirupsen logger, please ignore me :)

PR #6, opened 2019-04-04
Create metrics to track number of issued certs for openshift and openshiftapps

fahlmant:
As the operator-sdk doens't have an ootb way to add metrics, this is following the dedicated-admin-operators way to add metrics.

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -1,0 +1,65 @@
+// Copyright 2018 RedHat
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metrics
+
+import (
+	"net/http"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	// MetricsEndpoint is the port to export metrics on
+	MetricsEndpoint = ":8080"
+)
+
+var (
+	metricCertsIssuedInLastDayOpenshiftCom = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "certman_operator_certs_in_last_day_openshift_com",
+		Help: "Report how many certs have been issued for Openshift.com in the last day",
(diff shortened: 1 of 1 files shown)


pkg/metrics/metrics.go
```diff
@@ -0,0 +1,65 @@
...
+
+const (
+	// MetricsEndpoint is the port to export metrics on
+	MetricsEndpoint = ":8080"
+)
+
+var (
+	metricCertsIssuedInLastDayOpenshiftCom = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "certman_operator_certs_in_last_day_openshift_com",
+		Help: "Report how many certs have been issued for Openshift.com in the last day",
```
tparikh: Updated "last day" to "last 24 hours"

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,65 @@
...
+	RegisterMetrics()
+	http.Handle("/metrics", prometheus.Handler())
+	go http.ListenAndServe(MetricsEndpoint, nil)
+}
+
+// RegisterMetrics for the operator
+func RegisterMetrics() error {
+	err := prometheus.Register(metricCertsIssuedInLastDayOpenshiftCom)
+	err = prometheus.Register(metricCertsIssuedInLastDayOpenshiftAppsCom)
+	return err
```
fahlmant: This should probably be checked between each register

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	// MetricsEndpoint is the port to export metrics on
+	MetricsEndpoint = ":8080"
+)
+
+var (
+	metricCertsIssuedInLastDayOpenshiftCom = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
rogbas: If this metrics works more like a counter, I recommend using a Prometheus `CounterVec` instead of a gauge

rogbas: Then you would have to come up with a way to reset the counter at the end of each calendar day

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,65 @@
...
+		Name: "certman_operator_certs_in_last_day_openshift_com",
+		Help: "Report how many certs have been issued for Openshift.com in the last day",
+	}, []string{"name"})
+	metricCertsIssuedInLastDayOpenshiftAppsCom = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
rogbas: same counter vs gauge question here

pkg/metrics/metrics.go
tparikh: It is not a counter.

PR #7, opened 2019-04-11
Update Dockerflie add build scripts and add Makefile

jharrington22:
This PR adds build scripts required to deploy this operator via the app-sre pipeline.


jharrington22:
cc @jmelis

PR #8, opened 2019-04-12
Create metrics to track number of issued certs for openshift and openshiftapps

fahlmant:
This PR implements the logic for updating the metrics and exposes the metrics on operator startup


pkg/metrics/crt_sh.go
```diff
@@ -0,0 +1,180 @@
...
+	data := [][]string{}
+
+	table := tablewriter.NewWriter(os.Stdout)
+	table.SetHeader([]string{"Name", "NOT BEFORE", "NOT AFTER"})
+	table.SetRowLine(true) // Enable row line
+
+	db, err := sql.Open("postgres", getPsqlInfo())
+
+	if err != nil {
+		log.Fatalf("%v\n", err)
```
tparikh: don't use log.fatal

pkg/metrics/crt_sh.go
```diff
@@ -0,0 +1,180 @@
...
+	for rows.Next() {
+		var name string
+		var notBefore string
+		var notAfter string
+		var row []string
+
+		err = rows.Scan(&name, &notBefore, &notAfter)
+
+		if err != nil {
+			log.Fatalf("%v\n", err)
```
tparikh: don't use log.fatal

pkg/metrics/queries.go
```diff
@@ -0,0 +1,98 @@
+package metrics
+
+const (
```
tparikh: Do you need all these constants for metrics?

pkg/metrics/crt_sh.go
```diff
@@ -0,0 +1,180 @@
...
+
+	log.Printf("\nFollowing certificates issued by Let's Encrypt for %s domain are expiring in %d days\n", domain, durationDays)
+
+	table.Render()
+
+	return data
+}
+
+// GetListOfCertsIssued returns a list of certs issued for a given domain since durationDays number of days ago
+func GetListOfCertsIssued(domain string, durationDays int) [][]string {
```
tparikh: These functions are suited for command line but not for metrics. Do you really want them in operator?

PR #17, opened 2019-04-16
only make certificate request for RH managed cluster

tparikh:
This PR checks that the cluster deployment is for a Red Hat managed cluster before requesting certificates.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -97,12 +97,12 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
 	// Do not make certificate request if the cluster is not a Red Hat managed cluster.
 	if val, ok := cd.Labels["api.openshift.com/managed"]; ok {
 		if val != "true" {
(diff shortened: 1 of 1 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -94,6 +94,18 @@ func (r *ReconcileClusterDeployment) Reconcile(request reconcile.Request) (recon
 		return reconcile.Result{}, err
 	}
 
+	// Do not make certificate request if the cluster is not a Red Hat managed cluster.
+	if val, ok := cd.Labels["api.openshift.com/managed"]; ok {
+		if val != "true" {
+			reqLogger.Info("Not a managed cluster", "Namespace", request.Namespace, "Name", request.Name)
```
joelddiaz: here and in the other logger line below it adds the 'Namespace' and 'Name' fields to the log output, but the very first line in the Reconcile() func already set up those fields (with different names 'Request.Namespace' and 'Request.Name')

tparikh: updated

PR #18, opened 2019-04-16
Srep 1143 prep for ci pipeline

lisa:


config/config.go
```diff
@@ -0,0 +1,20 @@
+// Copyright 2018 RedHat
```
tparikh: 2019 :-)

lisa: haha! All the others probably say 2018 too!

PR #19, opened 2019-04-16
Unit test fixes

tparikh:
This PR adds the correct label to Cluster Deployments that are created for Red Hat managed clusters.

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -29,7 +29,8 @@ import (
 var log = logf.Log.WithName("controller_clusterdeployment")
 
 const (
-	controllerName = "clusterdeployment"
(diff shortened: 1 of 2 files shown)


tparikh:
@lisa FYI... This PR will fix the unit test failure you saw earlier

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -29,7 +29,8 @@ import (
 var log = logf.Log.WithName("controller_clusterdeployment")
 
 const (
-	controllerName = "clusterdeployment"
+	controllerName                = "clusterdeployment"
+	ClusterDeploymentManagedLabel = "api.openshift.com/managed"
```
joelddiaz: fyi, no need for it to be a capitalized/exported variable since the test code is part of the same package and see the private variables, but not a big deal

PR #35, opened 2019-05-13
Don't process cluster deployment until installed is set to true.

tparikh:
Only process a cluster deployment resource after OpenShift v4 cluster has been succesfully installed and "Installed" is set to true.

Resolves JIRA SREP-1263


pkg/controller/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -245,6 +245,9 @@ func testClusterDeployment() *hivev1alpha1.ClusterDeployment {
 				},
 			},
 		},
+		Status: hivev1alpha1.ClusterDeploymentStatus{
+			Installed: true,
```
fahlmant: Are you using this to create an object through the API, or just reading the struct in code?

PR #36, opened 2019-05-14
Read operator config from configmap

tparikh:
Operator can be made to use the Let's Encrypt staging endpoint using configuration. Default notification email address can also be set using the config map.

### DIFF
--- a/pkg/apis/certman/v1alpha1/certificaterequest_types.go
+++ b/pkg/apis/certman/v1alpha1/certificaterequest_types.go
@@ -40,16 +40,11 @@ type CertificateRequestSpec struct {
 	DnsNames []string `json:"dnsNames"`
 
 	// Let's Encrypt will use this to contact you about expiring certificates, and issues related to your account.
-	// +optional
 	Email string `json:"email"`
 
 	// Certificate renew before expiration duration in days.
 	// +optional
 	RenewBeforeDays int `json:"renewBeforeDays,omitempty"`
-
-	// Request Certificate from Let's Encrypt Staging Environment
-	// +optional
-	RequestTestCertificate bool `json:"requestTestCertificate,omitempty"`
 }
 
 type CertificateRequestCondition struct {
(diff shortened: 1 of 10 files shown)


jharrington22:
@tparikh so if the configmap doesn't exist the operator will use the staging env? Should we call that out in the README?

jharrington22:
@tparikh or maybe add a log line to indicate that its using staging?

pkg/apis/certman/v1alpha1/certificaterequest_types.go
```diff
@@ -40,16 +40,11 @@ type CertificateRequestSpec struct {
 	DnsNames []string `json:"dnsNames"`
 
 	// Let's Encrypt will use this to contact you about expiring certificates, and issues related to your account.
-	// +optional
 	Email string `json:"email"`
```
dak1n1: Hmm, the comment says this is optional, but the `Email` field will show up in json, even if it's unused, since it doesn't have  `omitempty`. Is this intentional? It is ok to have the field blank when it's marshalled into json? Like `Email: ""`

tparikh: With this commit, it is no longer optional. The diff is removing the `optional` annotation.

pkg/controller/controllerutils/configmap.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func GetDefaultNotificationEmailAddress(kubeClient client.Client) (string, error) {
+	cm, err := GetConfig(kubeClient)
+	if err != nil {
+		return "", err
+	}
+
+	if cm.Data["default_notification_email_address"] == "" {
+		return "", fmt.Errorf("")
```
rogbas: shouldn't we output an error message if email_address is blank?

tparikh: Seems like I missed adding text to that error. I will add error message.

PR #37, opened 2019-05-16
Deploy needed objects for prometheus metrics when the operator comes up

fahlmant:
This PR ensures that when the operator is deployed, all the proper objects are placed so prometheus can scrape the custom metrics.

-The operator creates a `certman-operator` service to expose the metrics over on deploy
-The operator creates a `certman-operator` servicemonitor to tell prometheus how to digest the metrics
-Removes the operator-sdk metrics service for now
-Includes a prometheus-k8s role for the `certman-operator` namespace that has permissions to get, list and watch services, pods and endpoints. 
-Includes a prometheus-k8s rolebinding to allow the `prometheus-k8s` serviceaccount from the `openshift-monitoring` namespace to use the aformentioned role.

When the operator and the role/rolebindings are deployed, the `certman-operator` namespace must be given the following label to get everything working: `openshift.io/cluster-monitoring="true"`

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -14,7 +14,8 @@ import (
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 	"github.com/operator-framework/operator-sdk/pkg/leader"
 	"github.com/operator-framework/operator-sdk/pkg/log/zap"
-	"github.com/operator-framework/operator-sdk/pkg/metrics"
+
+	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
 	"github.com/spf13/pflag"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
@@ -106,16 +107,30 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err := monitoringv1.AddToScheme(mgr.GetScheme()); err != nil {
+		log.Error(err, "error registering prometheus monitoring objects")
+		os.Exit(1)
+	}
+
 	// Setup all Controllers
 	if err := controller.AddToManager(mgr); err != nil {
 		log.Error(err, "")
 		os.Exit(1)
 	}
 
-	// Create Service object to expose the metrics port.
-	_, err = metrics.ExposeMetricsPort(ctx, metricsPort)
+	s, _ := operatormetrics.GenerateService(8080, "metrics")
+	sm := operatormetrics.GenerateServiceMonitor(s)
+	err = mgr.GetClient().Create(context.TODO(), s)
 	if err != nil {
-		log.Info(err.Error())
+		log.Error(err, "error creating metrics Service")
+	} else {
+		log.Info("Created Service")
+		err = mgr.GetClient().Create(context.TODO(), sm)
+		if err != nil {
+			log.Error(err, "error creating metrics ServiceMonitor")
+		} else {
+			log.Info("Created ServiceMonitor")
+		}
 	}
 
 	// Start metrics for prometheus
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -15,7 +15,7 @@ import (
 	"github.com/operator-framework/operator-sdk/pkg/leader"
 	"github.com/operator-framework/operator-sdk/pkg/log/zap"
 
-	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
+	routev1 "github.com/openshift/api/route/v1"
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
 	"github.com/spf13/pflag"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
@@ -107,7 +107,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	if err := monitoringv1.AddToScheme(mgr.GetScheme()); err != nil {
+	if err := routev1.AddToScheme(mgr.GetScheme()); err != nil {
 		log.Error(err, "error registering prometheus monitoring objects")
 		os.Exit(1)
 	}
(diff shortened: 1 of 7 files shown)


cmd/manager/main.go
```diff
@@ -106,17 +107,29 @@ func main() {
...
+	sm := operatormetrics.GenerateServiceMonitor(s)
+	err = mgr.GetClient().Create(context.TODO(), s)
+	if err != nil {
+		log.Error(err, "error creating metrics Service")
+	}
+	log.Info("Created Service")
+	err = mgr.GetClient().Create(context.TODO(), sm)
 	if err != nil {
-		log.Info(err.Error())
+		log.Error(err, "error creating metrics ServiceMonitor")
```
rogbas: AFAICS we are not exiting if the service/sm creation fails, I would exit and abort the operator startup if it fails

fahlmant: I couldn't decide if it should exit or not, happy to add it

rogbas: I think it makes sense to exit, to avoid having a metric-less operator running

tparikh: @fahlmant I would prefer to not exit if there is problem with registering metrics. Core functionality of the operator is not impacted when unable to export metrics.

cmd/manager/main.go
```diff
@@ -106,17 +107,29 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err := monitoringv1.AddToScheme(mgr.GetScheme()); err != nil {
+		log.Error(err, "error registering prometheus monitoring objects")
+		os.Exit(1)
```
tparikh: you can use log.Fatalf instead of log.errr and then os.exit

cmd/manager/main.go
```diff
@@ -106,17 +107,29 @@ func main() {
...
 
-	// Create Service object to expose the metrics port.
-	_, err = metrics.ExposeMetricsPort(ctx, metricsPort)
+	s, _ := operatormetrics.GenerateService(8080, "prom-metrics")
+	sm := operatormetrics.GenerateServiceMonitor(s)
+	err = mgr.GetClient().Create(context.TODO(), s)
+	if err != nil {
+		log.Error(err, "error creating metrics Service")
+	}
+	log.Info("Created Service")
```
tparikh: I think if there is an error above this line, then you will get log messages that say that "error creating metrics service" and "created service"

cmd/manager/main.go
```diff
@@ -106,17 +107,29 @@ func main() {
...
+	err = mgr.GetClient().Create(context.TODO(), sm)
```
tparikh: If service creation failed earlier, should we really continue with service monitor creation?

pkg/metrics/service.go
```diff
@@ -0,0 +1,92 @@
+// Copyright 2018 The Operator-SDK Authors
```
tparikh: 2019

cmd/manager/main.go
fahlmant: I'm using what was above in 106-107 that was generated from operator-sdk

cmd/manager/main.go
```diff
@@ -106,16 +107,32 @@ func main() {
...
+	s, _ := operatormetrics.GenerateService(8080, "prom-metrics")
+	sm := operatormetrics.GenerateServiceMonitor(s)
+	err = mgr.GetClient().Create(context.TODO(), s)
 	if err != nil {
-		log.Info(err.Error())
+		log.Error(err, "error creating metrics Service")
+	} else {
+		log.Info("Created Service")
+	}
+	if s != nil {
```
tparikh: error from creating service will not set `s` to nil unless I am missing something. You may have to move the logic inside  else block.

cmd/manager/main.go
```diff
@@ -106,16 +107,32 @@ func main() {
...
+		err = mgr.GetClient().Create(context.TODO(), sm)
+		if err != nil {
+			log.Error(err, "error creating metrics ServiceMonitor")
+		} else {
+			log.Info("Created ServiceMonitor")
+		}
 	}
 
 	// Start metrics for prometheus
```
tparikh: What happens if service or servicemonitor cannot be created? Should we even make the call to startmetrics?

cmd/manager/main.go
fahlmant: Ah I was looking at the GenerateService() function. One sec

cmd/manager/main.go
fahlmant: Yes.  We still want metrics to be available from the pod.

cmd/manager/main.go
```diff
@@ -106,16 +107,30 @@ func main() {
...
+
 	// Setup all Controllers
 	if err := controller.AddToManager(mgr); err != nil {
 		log.Error(err, "")
 		os.Exit(1)
 	}
 
-	// Create Service object to expose the metrics port.
-	_, err = metrics.ExposeMetricsPort(ctx, metricsPort)
+	s, _ := operatormetrics.GenerateService(8080, "metrics")
```
dak1n1: I think it would be valuable to do something with the error returned from `GenerateService` instead of using `_` to discard it. Specifically, this is where an error is being thrown in another operator I'm working on, and by default mine used `_`, which hid the error.

dak1n1: You can see mine here, if you're interested. It just lets the user know why the operator is crashing. https://github.com/openshift/configure-alertmanager-operator/pull/18/files

deploy/prometheus-k8s-role.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
dak1n1: According to our operator standards, this file should be in a directory called `manifests`. (I know this operator was made back before we had that standard, but if it's easy enough to `git mv` it, we may as well).

fahlmant: We have a JIRA card to track fixing this, so I think it will be done in a future PR.

fahlmant:
@tparikh Ping

PR #39, opened 2019-05-16
Certman Operator should watch all namespaces

tparikh:
As each ClusterDeployment resource is created in a separate namespace, certman operator needs to watch all namespaces.


fahlmant:
You will also need to alter some of the `deploy/` files:
https://github.com/operator-framework/operator-sdk/blob/master/doc/user-guide.md#operator-scope

PR #51, opened 2019-06-28
Fix the image versioning scheme to follow semver requirements

tparikh:
The current image version includes commit hash in it. If the commit hash starts with 0 (zero), operator image will not be deployed. See https://semver.org/#spec-item-2 and https://semver.org/#spec-item-9 for semver restrictions on leading zeros.

This commit adds build metadata (hash) after `+` as recommended in https://semver.org/#spec-item-10

Resolves Jira SREP-1692


tparikh:
@rogbas PTAL

PR #57, opened 2019-07-19
Error handling

c-e-brumm:
Pull request adds new functionality and lays ground work for additional error handling and reporting.

Error condition handled is when cluster name exceeds 100 characters and Let's Encrypt will not issue a cert. This pull request detects this error and updates the CertificateRequest status as 'Error' and gives the requests return body in the Status message.


tparikh:
@c-e-brumm As we will never be able to create certificates when we face this error, I would recommend that the check be put in the cluster deployment controller that creates the certificate request. Another possible workaround for this might be to create a domain name specifically to meet the CN length requirement even though this domain will not be used by anything in the cluster. Something like `osd.clusterid.env.openshiftapps.com`.

pkg/apis/certman/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -1,6 +1,6 @@
 // +build !ignore_autogenerated
 
-// Code generated by operator-sdk. DO NOT EDIT.
+// Code generated by deepcopy-gen. DO NOT EDIT.
```
fahlmant: Just curious, what version of operator-sdk are you using?

jewzaam:
> @c-e-brumm As we will never be able to successfully request certificates when we face this error, I would recommend that the check be put in the cluster deployment controller that creates the certificate request. Another possible workaround for this might be to create a domain name specifically to meet the CN length requirement even though this domain will not be used by anything in the cluster. Something like `osd.clusterid.env.openshiftapps.com`.

@tparikh that is a separate change.  What this PR addresses is a request is made for something too long, to bubble up the error and report an error in the certificate request.

tparikh:
@jewzaam I am not sure what we will gain by setting the status of `certificaterequest` to `Error`.  Certman Operator will never will be able to issue certificate for cluster than runs into this error. The only recourse in this scenario is to delete the cluster and request a new one with shorter name. The domain name length validation should either be done by Hive/UHC to prevent creation of cluster for which Certman Operator cannot issue certificate.

jewzaam:
@tparikh agree on all points.  Maybe I'm missing how we bubble up this error.  This change isn't to try issuing a new cert or to limit the name.  That should be handled externally.  This is just to get the right error surfaced so those external systems/people can make a decision on next steps.

pkg/apis/certman/v1alpha1/zz_generated.deepcopy.go
c-e-brumm: operator-sdk version: v0.8.1, commit: 33b3bfe10176f8647f5354516fff29dea42b6342

Yeah, I thought that was an odd diff too.

c-e-brumm:
If the ClusterDeployment status is already set to Installed = True then does it make sense to update the ClusterDeployment after it's reached that state? And if not then where else should this error be exposed?

jewzaam:
> If the ClusterDeployment status is already set to Installed = True then does it make sense to update the ClusterDeployment after it's reached that state? And if not then where else should this error be exposed?

Don't update ClusterDeployment, this operator doesn't own it.  Long term we have plans to use ClusterOperator CRs to report status both on the Hive cluster and on individual OSD clusters.  This would be used to aggregate status beyond just "installed" which can then reflect state day-2 operations to the end customer.

c-e-brumm:
Agreed, editing a CR you don't own is dangerous. I don't recommend it, just pointing out that it isn't a good option. Doubly so if it already was set to a final state like this case.

So do we bubble up Let's Encrypt errors to the CertificateRequest CR and wait for the next stage to be built? Or is there a corrective path to make to this PR/feature patch?

tparikh:
@c-e-brumm There is no next stage. No other component will react to the `certificaterequest` in `error` state. We are catching the error too late in the cluster provisioning process.

jewzaam:
@tparikh bubbling it up into the certificaterequest status will let someone triage the problem a bit easier.  The code using this for our use case already is updated to not hit this case again.  Doesn't mean this change isn't valid.

tparikh:
@jewzaam Who is monitoring these errors? What corrective action can they take? I am not against setting the status to error. I don't think code change in this PR solves the actual problem. I have no objection to merging the changes as long as everyone knows that the actual problem still needs to be addressed.

/hold cancel

tparikh:
I don't see any issue with the code itself.

/lgtm

PR #58, opened 2019-07-24
using operator-custom-metrics library

ninataneja:
Gets rid of pkg/metrics/service.go and the logic for bringing up the prometheus metric server. Moves to using github.com/openshift/operator-custom-metrics pkg to register and expose metrics.


tparikh:
@fahlmant @c-e-brumm  Can you both plese review this PR

/assign @fahlmant

PR #60, opened 2019-08-01
Added metric for runtime of issue certificate function

ninataneja:
Added metric to record the time (in seconds) for the certificate to be created (the time of the issue certificate function). This metric will allow us to create an alert if this time is over 15 minutes.

### DIFF
--- a/pkg/localmetrics/localmetrics.go
+++ b/pkg/localmetrics/localmetrics.go
@@ -41,6 +41,10 @@ var (
 		Name: "certman_operator_duplicate_certs_in_last_week",
 		Help: "Report how many certs have had duplicate issues",
 	}, []string{"name"})
+	MetricIssueCertificateDuration = prometheus.NewSummary(prometheus.SummaryOpts{
+		Name: "certman_operator_certificate_creation_duration",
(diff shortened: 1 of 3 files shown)


pkg/localmetrics/localmetrics.go
```diff
@@ -41,13 +41,18 @@ var (
 		Name: "certman_operator_duplicate_certs_in_last_week",
 		Help: "Report how many certs have had duplicate issues",
 	}, []string{"name"})
+	MetricIssueCertificateDuration = prometheus.NewSummary(prometheus.SummaryOpts{
+		Name: "certman_operator_certificate_creation_duration",
```
tparikh: replace `creation` to `issue`

pkg/localmetrics/localmetrics.go
```diff
@@ -88,3 +93,7 @@ func UpdateMetrics(hour int) {
 		UpdateDuplicateCertsIssuedInLastWeek()
 	}
 }
+
+func UpdateCertificateCreationDurationMetric(time time.Duration) {
```
tparikh: Replace `Creation` with `Issue`

Issue #30, opened 2019-05-06
Remove the "staging" field from CRD


tparikh:
Let's Encrypt staging/production endpoint should be passed to operator as configuration parameters. There is no real case where we would request certificates from Let's Encrypt staging API endpoint in production environment.

dofinn:
Resolved in: https://github.com/openshift/certman-operator/pull/69

PR #61, opened 2019-08-05
Extracting Let's Encrypt to a client interface

c-e-brumm:
This will help with dependency injection and creating mock clients.


pkg/leclient/lets_encrypt.go
```diff
@@ -0,0 +1,199 @@
...
+		Client.Client, err = acme.NewClient(acme.LetsEncryptStaging)
+		return Client, err
+	}
+	Client.Client, err = acme.NewClient(acme.LetsEncryptProduction)
+	return Client, err
+}
+
+func getLetsEncryptAccountPrivateKey(kubeClient client.Client, staging bool) (privateKey crypto.Signer, err error) {
+
+	secretName := LetsEncryptProductionAccountSecretName
```
cblecker: Question: Do we expect an install of this operator to have both staging and prod? Is there a reason to have two different secrets in the same NS?

pkg/controller/certificaterequest/issue_certificate.go
```diff
@@ -53,80 +54,59 @@ func (r *ReconcileCertificateRequest) IssueCertificate(reqLogger logr.Logger, cr
...
 		}
 
-		reqLogger.Info("authorization url", "authorization.URL", authorization.URL)
-
-		domain := authorization.Identifier.Value
-
-		challenge, ok := authorization.ChallengeMap[acme.ChallengeTypeDNS01]
-		if !ok {
+		domain := leClient.GetAuthorizationIndentifier()
+		leClient.SetChallengeType()
```
cblecker: Uncaught error handling

pkg/controller/certificaterequest/issue_certificate.go
```diff
@@ -53,80 +54,59 @@ func (r *ReconcileCertificateRequest) IssueCertificate(reqLogger logr.Logger, cr
...
+		if err != nil {
 			return fmt.Errorf("cloud not find DNS challenge authorization")
 		}
 
-		encodeDNS01KeyAuthorization := acme.EncodeDNS01KeyAuthorization(challenge.KeyAuthorization)
+		DNS01KeyAuthorization := leClient.GetDNS01KeyAuthorization()
```
cblecker: Is it possible for this response to be `nil`?

pkg/controller/certificaterequest/issue_certificate.go
```diff
@@ -53,80 +54,59 @@ func (r *ReconcileCertificateRequest) IssueCertificate(reqLogger logr.Logger, cr
...
 			return err
 		}
 
-		reqLogger.Info("authorization url", "authorization.URL", authorization.URL)
-
-		domain := authorization.Identifier.Value
-
-		challenge, ok := authorization.ChallengeMap[acme.ChallengeTypeDNS01]
-		if !ok {
+		domain := leClient.GetAuthorizationIndentifier()
```
cblecker: Is it possible for this response to be `nil`?

pkg/controller/certificaterequest/issue_certificate.go
```diff
@@ -53,80 +54,59 @@ func (r *ReconcileCertificateRequest) IssueCertificate(reqLogger logr.Logger, cr
...
 		certDomains = append(certDomains, domain)
-		ids = append(ids, acme.Identifier{Type: "dns", Value: domain})
 	}
 
-	letsEncryptOrder, err := letsEncryptClient.NewOrder(letsEncryptAccount, ids)
+	err = leClient.CreateOrder(cr.Spec.DnsNames)
 	if err != nil {
 		return err
 	}
+	URL, _ := leClient.GetOrderURL()
```
cblecker: Why are we discarding the err here?

pkg/controller/certificaterequest/update_status.go
```diff
@@ -42,8 +42,7 @@ func (r *ReconcileCertificateRequest) updateStatus(reqLogger logr.Logger, cr *ce
 			cr.Status.IssuerName != certificate.Issuer.CommonName ||
 			cr.Status.NotBefore != certificate.NotBefore.String() ||
 			cr.Status.NotAfter != certificate.NotAfter.String() ||
-			cr.Status.SerialNumber != certificate.SerialNumber.String() ||
-			cr.Status.Status != "Success" {
```
cblecker: This seems unrelated to the rest of the PR. Was this intended?

pkg/leclient/lets_encrypt.go
```diff
@@ -0,0 +1,199 @@
...
+		ids = append(ids, acme.Identifier{Type: "dns", Value: domain})
+	}
+	c.Order, err = c.Client.NewOrder(c.Account, ids)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (c *ACMEClient) GetAccount(kubeClient client.Client, staging bool, namespace string) (err error) {
```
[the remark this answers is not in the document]
c-e-brumm: This is legacy logic that I wasn't trying to remove in this PR. But you're right, there is absolutely no need for staging vs production endpoints in the operator logic. It needs to just be config data supplied to the operator (endpoint, account name, keys, etc). That will be fixed in a future PR.

pkg/leclient/constants.go
```diff
@@ -0,0 +1,35 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package leclient
+
+const (
```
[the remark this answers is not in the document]
c-e-brumm: No, this is actually lifted straight from the package certificaterequest constants, where it originated, to get the pieces I needed. I forgot to curate these, good catch.

pkg/leclient/lets_encrypt.go
c-e-brumm: No, it will never do both, which is why it needs to be streamlined out of the code. That's coming in another story/PR.

pkg/controller/certificaterequest/issue_certificate.go
c-e-brumm: Neither SetChallengeType or GetAuthorizationIndentifier should be able to fail but I've added error outputs and stanzas just in case in the future they can. Or in case future implementations can fail.

pkg/controller/certificaterequest/update_status.go
c-e-brumm: Not related but I accidentally committed bug fix to my feature branch. This removes always catching a non "Status: Success" and setting it to "Success".

pkg/controller/certificaterequest/issue_certificate.go
c-e-brumm: Do not believe so. But I'll handling just in case.

pkg/controller/certificaterequest/issue_certificate.go
c-e-brumm: No, but check added in case.

pkg/controller/certificaterequest/issue_certificate.go
c-e-brumm: The left side is unchanged from before. I've fixed this and added handling.

cblecker:
@app-sre-bot test this

cblecker:
[test]

cblecker:
/lgtm
/approve
/hold

Please feel free to remove hold when you are ready to merge.

pkg/leclient/lets_encrypt.go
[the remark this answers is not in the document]
dofinn: Are tasks like this getting managed via Jira?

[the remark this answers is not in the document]
dofinn: https://github.com/openshift/certman-operator/issues/30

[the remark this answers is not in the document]
c-e-brumm: All certman-operator stories, https://jira.coreos.com/browse/SDE-109
This item specifically, https://jira.coreos.com/browse/SREP-1802

PR #62, opened 2019-08-05
Added metric for days until certificate expiry

ninataneja:
Added a metric for the number of days a certificate is valid for (until expiry), with labels for the certificate serial number and the corresponding cluster URL. This metric will allow us to create an alert for when the certificate expiry date is within a certain # of days (for example, send an alert if a certificate is expiring in the next 15 days)

### DIFF
--- a/pkg/controller/certificaterequest/renew_certificate.go
+++ b/pkg/controller/certificaterequest/renew_certificate.go
@@ -60,10 +60,6 @@ func (r *ReconcileCertificateRequest) ShouldRenewOrReIssue(reqLogger logr.Logger
 		timeDiff := notAfter.Sub(currentTime)
 		daysCertificateValidFor := int(timeDiff.Hours() / 24)
 		shouldRenew := daysCertificateValidFor <= renewBeforeDays
-		certSerialNum := certificate.SerialNumber
-
-		localmetrics.UpdateCertExpiryDateMetric(cr.Spec.WebConsoleURL, certSerialNum, daysCertificateValidFor)
-
 		localmetrics.UpdateCertExpiryDateMetric(cr.Name, cr.Spec.ACMEDNSDomain, daysCertificateValidFor)
 
 		if shouldRenew {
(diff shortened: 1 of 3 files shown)


pkg/controller/certificaterequest/renew_certificate.go
```diff
@@ -59,6 +60,9 @@ func (r *ReconcileCertificateRequest) ShouldRenewOrReIssue(reqLogger logr.Logger
 		timeDiff := notAfter.Sub(currentTime)
 		daysCertificateValidFor := int(timeDiff.Hours() / 24)
 		shouldRenew := daysCertificateValidFor <= renewBeforeDays
+		certSerialNum := certificate.SerialNumber
+
+		localmetrics.UpdateCertExpiryDateMetric(cr.Spec.WebConsoleURL, certSerialNum, daysCertificateValidFor)
```
tparikh: The dimensions on this metric don't make much sense. You should log following dimensions for each metrics

- cr.Name
- cr.Spec.ACMEDNSDomain
- cert_expiry_month
- cert_expiry_day
- cert_expiry_year
- cert_expiring_in_days

Webconsole URL and Serial number are not needed.

fahlmant: I disagree. Labels should only be used to categorize metrics, not have data in them.

tparikh: The dimensions used here currently don't help in any way with the metrics.

fahlmant: They categorize the metric by signifying which cert/cluster is expiring.

tparikh: @fahlmant Yes and the dimensions added in the initial code do not allow for that. We do not want console url. We identify cluster using ACMEBaseDomain.

fahlmant: Okay, by why cert expiry year, day, month and expiring in days? That will create a new metric for every day

tparikh: OK I have updated the dimension.

fahlmant: LGTM :+1:

ninataneja:
found potential bug/issue with Prometheus gauge labels, working on figuring it out

c-e-brumm:
Way too old, will review if needed.

PR #64, opened 2019-08-16
[WIP] Add checks and exponential backoff to remote API calls

dak1n1:
* Created new package `sleep` for exponential backoff sleep, which can be used in other projects.
* Check Let's Encrypt and AWS credentials before attempting to use them.
* Back off after failed API requests to Let's Encrypt and AWS.

https://jira.coreos.com/browse/SREP-1949


pkg/controller/certificaterequest/certificaterequest_controller.go
```diff
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/aws/aws-sdk-go/service/route53"
```
c-e-brumm: Instead we should import the local package certman-operator/awsclient instead of the proper route53 package. This will allow us to mock the client for unit tests. The awsclient already implements the ListHostedZones function so it should be a drop in replacement.

pkg/leclient/utils.go
```diff
@@ -36,3 +41,37 @@ func GetSecret(kubeClient client.Client, secretName, namespace string) (*corev1.
...
+	for i, condition := range cr.Status.Conditions {
+		if string(condition.Type) == queryType {
+			failCount, err := strconv.Atoi(string(condition.Status))
+			if err != nil {
+				return err
+			}
+			// Sleep an exponential number of seconds after each API call failure.
+			// For example: 2 seconds, 4 seconds, 16 seconds, 32 seconds...
+			seconds := 200 * time.Millisecond
+			sleeptime := seconds << uint(failCount)
```
c-e-brumm: Let's also put a cap in it. Something like 
`sleeptime = math.Min(100, sleeptime)`
or whatever number we choose for a max backoff length.

pkg/controller/certificaterequest/certificaterequest_controller.go
```diff
@@ -145,18 +146,41 @@ func (r *ReconcileCertificateRequest) Reconcile(request reconcile.Request) (reco
...
+	// Ensure that platform Secret exists.
+	awscreds := &corev1.Secret{}
+	credserr := r.client.Get(context.TODO(), types.NamespacedName{Name: platformSecretName, Namespace: request.Namespace}, awscreds)
+	if credserr != nil {
+		reqLogger.Info("platformSecrets were not found. Unable to search for certificates in cloud provider platform")
+		return reconcile.Result{}, credserr
+	}
 
-	err = r.client.Get(context.TODO(), types.NamespacedName{Name: certificateSecret.Name, Namespace: certificateSecret.Namespace}, found)
+	// Ensure that platform Secret can authenticate to AWS.
```
c-e-brumm: I think it might be best if this logic block was moved behind the awsclient and abstracted away. We could call it awsclient.TestAuth() or something like that.

cmd/manager/main.go
```diff
@@ -129,7 +129,6 @@ func main() {
 	// Configure metrics. If it errors, log the error but continue.
 	if err := metrics.ConfigureMetrics(context.TODO(), *metricsServer); err != nil {
 		log.Error(err, "Failed to configure Metrics")
-		os.Exit(1)
```
dak1n1: I removed this so the operator can be run in minikube. It means that metrics failures will be non-fatal, which complies with the team's standards for metrics. Also makes behavior match the comment above it.

jharrington22:
@c-e-brumm @dak1n1 I have a quick question regarding the exponential backoff here. Would it be better (or maybe its not possible without looking closer) to make this reconcile loop more idempotent and `return reconcile.Result{}, err` from the reconcile loop when you need exponential backoff as it comes for free when doing that.

jharrington22:
> @c-e-brumm @dak1n1 I have a quick question regarding the exponential backoff here. Would it be better (or maybe its not possible without looking closer) to make this reconcile loop more idempotent and `return reconcile.Result{}, err` from the reconcile loop when you need exponential backoff? as it comes for free when doing that.

The retry would be very fast at the start so it might not be useful, however sleeping within the controller will block any other events from being processed during that time, It might be worth looking at requeuing the request in the future using your same exponential backoff logic. You can requeue a request in the future with `reconcile.Result{RequeueAfter: time.Duration(10) * time.Second}, nil` for example.

cmd/manager/main.go
dak1n1: I thought we had a discussion about this as a team, and while I agreed with you that it's better to have metrics failures be fatal, we were supposed to make them non-fatal? I'll update the operator guidelines to clarify this, and put the `os.Exit(1)` back.

fahlmant: We had a more recent discussion where it came up again and the consensus changed to making them fatal. I put this line in PRs across all the hive operators and they now all have it. Thanks for updating the guidelines doc, I forget that it exists sometimes.

PR #68, opened 2019-08-22
Updating certs with new DNSNames

c-e-brumm:
When a CertificateRequest reconcile has a DNSName not in the current cert request a new cert.


dofinn:
/lgtm 
confirming that renewing and first issuing take the same code path?

PR #69, opened 2019-08-23
Simplify let's encrypt staging/production env setup

wanghaoran1988:
After this:
1.  we do not need set lets_encrypt_environment=staging in the certman-operator configmap, we will get the lets encrypt directory url according to the account-url.
2. we need create "lets-encrypt-account" secret instead of "lets-encrypt-account-staging" and "lets-encrypt-account-production", but for backward compatability, if "let's-encrypt-account" secret is not there, it will try to use "lets-enctrypt-account-production", then "lets-encrypt-account-staging".

### DIFF
--- a/pkg/controller/certificaterequest/issue_certificate.go
+++ b/pkg/controller/certificaterequest/issue_certificate.go
@@ -30,7 +30,6 @@ import (
 
 	"github.com/openshift/certman-operator/config"
 	certmanv1alpha1 "github.com/openshift/certman-operator/pkg/apis/certman/v1alpha1"
-	"github.com/openshift/certman-operator/pkg/controller/controllerutils"
 	"github.com/openshift/certman-operator/pkg/leclient"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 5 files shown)


dofinn:
@wanghaoran1988  we seem to be replacing logic with duplicate blocks of code, any thoughts around re-factoring? to get the client once and pass it around? (`issue_certificate.go` and `revoke_certificate.go`)

wanghaoran1988:
> ```
> durl = ""
> 	if strings.Contains(acme.LetsEncryptStaging, u.Host) {
> 		durl = acme.LetsEncryptStaging
> 	} else if strings.Contains(acme.LetsEncryptProduction, u.Host) {
> 		durl = acme.LetsEncryptStaging
> 	} else {
> 		return "", errors.New("cannot found let's encrypt directory url.")
> ```
> 
> We are assigning `durl` the same Var regardless of the if/else logic.

Good catch

wanghaoran1988:
> @wanghaoran1988 we seem to be replacing logic with duplicate blocks of code, any thoughts around re-factoring? to get the client once and pass it around? (`issue_certificate.go` and `revoke_certificate.go`)

We can create a seprate issue for this, I am not clear whether we should have only one "letsencrypt client" instance or create a new client every time.

PR #70, opened 2019-08-26
Update clusterDeployment status as needed

wanghaoran1988:
https://jira.coreos.com/browse/SREP-1192
/cc @c-e-brumm @dofinn


wanghaoran1988:
Updated

PR #72, opened 2019-08-30
[SREP-1954] Added comments to all uncommented functions

dofinn:
exploring package inheritance and usage


c-e-brumm:
Huge +1 on adding the licenses, thank you for that. Great job on the comments this far.

PR #85, opened 2019-10-11
Adding deletion for test access record after successful write.

c-e-brumm:
After the testing for proper write access for route53 we test our deletion abilities and clear up the record.


cblecker:
[test]

PR #90, opened 2019-10-22
Delete all acme challenge records

c-e-brumm:
After completing an acme challenge all acme challenge records for the same hosted zone are deleted. This will ensure that no records are left behind in a zone. And if records are left behind from before they will be cleaned up when the zone is used next.

### DIFF
--- a/pkg/awsclient/client.go
+++ b/pkg/awsclient/client.go
@@ -79,7 +79,7 @@ func (c *awsClient) ListResourceRecordSets(input *route53.ListResourceRecordSets
 	return c.route53Client.ListResourceRecordSets(input)
 }
 
-// Function for finding a hostedzone when given an aws client and a domain string
+// SearchForHostedZone finds a hostedzone when given an aws client and a domain string
 // Returns a hostedzone object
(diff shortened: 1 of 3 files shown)


pkg/awsclient/client.go
```diff
@@ -78,6 +79,52 @@ func (c *awsClient) ListResourceRecordSets(input *route53.ListResourceRecordSets
...
+		HostedZoneId: &hostedZone,
+	}
+	return input
+}
+
+// CreateR53TXTRecordChange creates an route53 Change object for a TXT record with a given name
+// and a given action to take.
+func CreateR53TXTRecordChange(name *string, action string, value *string) route53.Change {
+	change := route53.Change{
+		Action: aws.String(route53.ChangeActionDelete),
```
dak1n1: I see `action` passed in, but not used. Should it go on this line instead of `ChangeActionDelete`?

c-e-brumm: Good catch. It looks like I forgot to write the switch/case statement for setting dynamic actions.

c-e-brumm: Updated, now checks the action passed and sets the correct r53 action.

PR #94, opened 2019-10-30
Add GCP support

mjudeikis:
Add GCP support and refactor interface 

In the PR:
* revendor hive api
* revendor operator SDK
* vendor GCP DNS API
* move clients to more abstract subpackage
* make certman to be orientated around the new interface for LetsEncrypt

PR #99, opened 2019-12-04
Revert "Add GCP support"

c-e-brumm:
Reverts openshift/certman-operator#94


dak1n1:
@mjudeikis FYI, we had to revert. I left a comment on the original PR where I think the problem is.

PR #102, opened 2019-12-06
Revert "Revert "Revert "Add GCP support"""

jharrington22:
Reverts openshift/certman-operator#101

This PR is to Revert the GCP changes which introduce two issues

1. The `certman_v1alpha1_certificaterequest_crd.yaml` is invalid which is breaking the operator install using OLM
2. This PR introduces modules however the dep dependdencies `Gopkg.lock` `Gopkg.toml` and the `/Vendor` directory have not been removed. We should follow instructions here: https://github.com/operator-framework/operator-sdk/blob/master/doc/migration/version-upgrade-guide.md

My opinion would be to remove the GCP changes from this PR and create a separate PR with those changes and not upgrade to Go modules or operator-sdk 0.8.2. In a second PR do the work to upgrade to Go module support and operator-sdk version.

This PR will restore the certman-operator in staging.

cc/ @jfchevrette Thanks for helping debug this issue.


dak1n1:
@mjudeikis FYI, we're still working through the OLM issues. It's specific to the deployment pipeline. I'll let you know when we're able to merge.

PR #107, opened 2019-12-19
New feature, additional control plane record added to certificate

c-e-brumm:
This will allow Certman Operator owners to specify and additional hostname of the same domain to be added to the certificate of the control plane.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,24 @@ oc create -f deploy/operator.yaml
 
 `certman_operator_duplicate_certs_in_last_week` reports how many certs have had duplication issues.
 
+## Additional record for control plane certificate
+
+Certman Operator always creates a certificate for the control plane for the clusters Hive builds. By passing a string into the pod as an environment variable named `EXTRA_RECORD` Certman Operator can add an additional record to the SAN of the certificate for the API servers. This string should be the short hostname without the domain. The record will use the same domain as the rest of the cluster for this new record.
+Example
+```
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: certman-operator
+spec:
+  template:
+    spec:
+    ...
+      env:
+      - name: EXTRA_RECORD
+        value: "myapi"
+```
+The example will add `myapi.<clustername>.<clusterdomain>` to the certificate of the control plane.
 ## License
 
 Certman Operator is licensed under Apache 2.0 license. See the [LICENSE](LICENSE) file for details.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -149,6 +149,24 @@ oc create -f deploy/operator.yaml
...
+  name: certman-operator
+spec:
+  template:
+    spec:
+    ...
+      env:
+      - name: EXTRA_RECORD
+        value: "myapi"
+```
+The example will add `myapi.<clustername>.<clusterdomain>` to the certificate of the control plane.
```
dak1n1: This example is really helpful!

pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -334,11 +335,17 @@ func getDomainsForCertBundle(cb hivev1alpha1.CertificateBundleSpec, cd *hivev1al
 	domains := []string{}
 	dLogger := logger.WithValues("CertificateBundle", cb.Name)
 
-	// first check for the special-case default control plane reference
+	// first check for the special-case two default control plane references
```
dak1n1: The code looks good, but the comment doesn't make much sense to me. (I know it's not originally your comment, but anything we can do to clear it up might be helpful.)

Maybe something like `// First check if CertificateBundleSpec.Name matches the default control plane name`.

Also, since that's what this block is doing, does it make sense to check for the env var inside this block? That is, do we really need the default ServingCertificate to equal the CertificateBundleSpec name before appending the extra record? What if the default Serving Cert is not equal to the CertificateBundleSpec?

Sorry for the inconvenience. I'm just trying to understand this fully.

c-e-brumm: The reason for this being in the if block is that this extra record is just an alternate name for the cluster api. Using it you can get a cluster with a cert for the cluster api of `api.cluster.domain` and `custom.cluster.domain`. But if we put it out of the if block it would be applied to ever other cert for the cluster.

Having a cluster with multiple certificates for custom ingress objects is not common but is supported by Hive. By default Hive for OSD asks for certs with `api.cluster`, `*.apps.cluster`. But the cluster deployment object may ask for additional certs and domain records which can be applied to Ingresses. Placing this option inside the if statements prevents the new option from being applied to these certs also.

That comment could totally be improved. +1 on that.

wanghaoran1988:
So, IIUC, the operator deployed with that env, all the controle plane certs will have a SAN added?

c-e-brumm:
Yes @wanghaoran1988, that's correct. We'll be making use of this in the future.

PR #116, opened 2020-02-20
Set AWS region from new CR property

sedroche:
Ticket: https://issues.redhat.com/browse/OSD-2789

Region is currently hard-coded. This change moves to using the ClusterDeployment, `Spec.Platform.aws.region` to determine the `CertificateRequest` region.


deploy/crds/certman_v1alpha1_certificaterequest_crd.yaml
```diff
@@ -95,12 +95,16 @@ spec:
               description: WebConsoleURL is the URL for the cluster's web console
                 UI.
               type: string
+            region:
```
wanghaoran1988: This does not lgtm, has a aws specific config in the top level doesn't make sense to all the other cloud provider, we may consider add the region info as an annotation in the CR object, or add another struct for platform specific, something like:

```
platform:
    aws:
      region: <region_name>
      <other platform related>
```

sedroche: That makes sense @wanghaoran1988 
I'd prefer to keep it as a property on the CRD, I see there is an existing `platformSecrets` property which could be updated to just `platform` and could contain the region info too
e.g.

```.yaml
platform:
  aws:
    credentials:
    region:


platform:
  gcp:
    credentials:
```

wdyt?

wanghaoran1988: Remove the duplicate "platform" ?
```
platform:
  aws:
    credentials:
    region:
  gcp:
    credentials:
```

sedroche: @wanghaoran1988 Yes, my mistake. Exactly that

c-e-brumm:
Looks good but I'm hesitant to merge late on Friday since there is a CRD change in the PR. Unknown if CD for int/stg will update the CRD correctly so waiting for a non-weekend day.

PR #125, opened 2020-03-24
Don't hard error on NXDOMAIN

cblecker:
Right now, if we get an NXDOMAIN (domain doesn't exist) when checking for propagation, we throw a hard error, causing us to retry the whole validation process.

In the case of a new cluster, where DNS propagation takes awhile, NXDOMAIN is expected initially. If we hard error, then we create a whole new DNS record causing the validation to never succeed.


cblecker:
/assign @c-e-brumm 
/hold

Holding as I want review/opinions if the third commit is helpful. 1 and 2 are necessary, but the 3rd may not be needed.

PR #127, opened 2020-03-25
Update certman to go1.13/operator-sdk 0.16

cblecker:
This updates certman to build on go1.13, and it's dependencies to be based off of operator-sdk 0.16.

As a side effect, it also moves it to use go modules, and removes the vendor directory.


cmd/manager/main.go
```diff
@@ -79,56 +87,61 @@ func start() error {
...
 	if err != nil {
 		log.Error(err, "")
-		return err
+		os.Exit(1)
+	}
+
+	// Set default manager options
+	options := manager.Options{
+		Namespace:          namespace,
+		MetricsBindAddress: fmt.Sprintf("%s:%d", metricsHost, metricsPort),
```
c-e-brumm, line 100: Setting up the sdk default Metrics here conflicts with our custom metrics server that's defined below. Removing line 109 allows the operator to run as expected in test environments.

cblecker:
coverage will fail until https://github.com/openshift/release/pull/7896 goes in

cblecker:
/hold cancel
This one I think is ready, @c-e-brumm

PR #131, opened 2020-04-06
OSD-3261 verify credential secrets are present

yithian:


c-e-brumm:
Tests pass, looks good.

PR #133, opened 2020-04-08
Moved strings to constants

dofinn:


pkg/clients/types/types.go
```diff
@@ -1,6 +1,8 @@
 package types
 
 const (
-	AcmeChallengeSubDomain   = "_acme-challenge"
-	WriteValidationSubDomain = "_certman_access_test"
+	AcmeChallengeSubDomain          = "_acme-challenge"
+	WriteValidationSubDomain        = "_certman_access_test"
+	DefaultNotificationEmailAddress = "default_notification_email_address"
+	LetsEncryptEnvironment          = "lets_encrypt_environment"
```
2uasimojo, line 7: I notice this one is only used in test.

PR #137, opened 2020-04-30
 cloudflare: Honor negative caching TTL

mbarnes:
This is split into 3 commits:

1. `Fix misspelling in variable names` is trivial and low risk
2. `Restructure verification code` is preparation so I can fix the bug
    _This is where a regression would be most likely._
3. `Honor negative caching TTL` is the actual bug fix

### DIFF
--- a/pkg/controller/certificaterequest/cloudflare.go
+++ b/pkg/controller/certificaterequest/cloudflare.go
@@ -49,23 +49,53 @@ type CloudflareResponse struct {
 	CD        bool                 `json:"CD"`
 	Questions []CloudflareQuestion `json:"Question"`
 	Answers   []CloudflareAnswer   `json:"Answer"`
+	Authority []CloudflareAnswer   `json:"Authority"`
 }
 
 // VerifyDnsResourceRecordUpdate verifies the presence of a TXT record with Cloudflare DNS.
 func VerifyDnsResourceRecordUpdate(reqLogger logr.Logger, fqdn string, txtValue string) bool {
+	var negativeCacheTTL int
+
 	for attempt := 1; attempt < maxAttemptsForDnsPropagationCheck; attempt++ {
 		var err error
 
+		// Sleep before querying Cloudflare DNS.  If the previous attempt returned
+		// a negative cache result, honor its TTL (within reason).  Otherwise wait
+		// for a predetermined duration.
 		sleepDuration := waitTimePeriodDnsPropagationCheck
+		if attempt > 1 && negativeCacheTTL > 0 {
+			// maxNegativeCacheTTL determines what is "reasonable".
+			// If the SOA TTL exceeds this, give up immediately.
+			if negativeCacheTTL > maxNegativeCacheTTL {
+				reqLogger.Info("negative cache TTL is too long; giving up")
+				return false
+			}
+			// If the TTL is shorter than the default wait time, disregard.
+			if negativeCacheTTL > sleepDuration {
+				sleepDuration = negativeCacheTTL
+			}
+		}
 
 		reqLogger.Info(fmt.Sprintf("attempt %v to verify resource record %v has been updated with value %v", attempt, fqdn, txtValue))
 
 		reqLogger.Info(fmt.Sprintf("will query DNS in %v seconds", sleepDuration))
 		time.Sleep(time.Duration(sleepDuration) * time.Second)
 
+		negativeCacheTTL = 0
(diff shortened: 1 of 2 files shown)


pkg/controller/certificaterequest/cloudflare.go
```diff
@@ -49,59 +49,84 @@ type CloudflareResponse struct {
...
+			reqLogger.Info(fmt.Sprintf("got a negative cache response with a TTL of %v seconds", ttl))
+			// maxNegativeCacheTTL determines what is "reasonable".
+			// If the SOA TTL exceeds this, give up immediately.
+			if sleepDuration > maxNegativeCacheTTL {
+				reqLogger.Info("negative cache TTL is too long; giving up")
+				return false
+			}
+			// If the TTL is shorter than the default wait time, disregard.
+			if ttl > sleepDuration {
+				sleepDuration = ttl
```
cblecker, line 80: Should we maybe add 5 extra seconds just to make sure thing things have expired?

pkg/controller/certificaterequest/cloudflare.go
```diff
@@ -49,59 +49,84 @@ type CloudflareResponse struct {
...
-	return false
-}
+		reqLogger.Info(fmt.Sprintf("will query DNS in %v seconds", sleepDuration))
+		time.Sleep(time.Duration(sleepDuration) * time.Second)
 
-// verifyDnsResourceRecordUpdate will wait the waitTimePeriodDnsPropogationCheck
-// and then check if the DNS changes have propagated
-func verifyDnsResourceRecordUpdate(reqLogger logr.Logger, fqdn string, txtValue string) bool {
-	reqLogger.Info(fmt.Sprintf("will query DNS in %v seconds", waitTimePeriodDnsPropogationCheck))
+		response = new(CloudflareResponse)
```
cblecker, line 90: Do we need to do this each time? Or can we just set `response := &CloudflareResponse{}` at the top, and overwrite it each time we do `FetchResourceRecordUsingCloudflareDNS`?

pkg/controller/certificaterequest/cloudflare.go
```diff
@@ -49,59 +49,84 @@ type CloudflareResponse struct {
...
-	return dnsChangesPropogated
+	errMsg := fmt.Sprintf("unable to verify that resource record %v has been updated to value %v after %v attempts.", fqdn, txtValue, maxAttemptsForDnsPropagationCheck)
+	reqLogger.Error(errors.New(errMsg), errMsg)
+	return false
 }
 
-// ValidateResourceRecordUpdatesUsingCloudflareDNS contacts cloudflareDnsOverHttpsEndpoint and validates the json response.
-func ValidateResourceRecordUpdatesUsingCloudflareDNS(reqLogger logr.Logger, name string, value string) (bool, error) {
+// FetchResourceRecordUsingCloudflareDNS contacts cloudflareDnsOverHttpsEndpoint and returns the json response.
+func FetchResourceRecordUsingCloudflareDNS(reqLogger logr.Logger, name string, outResponse *CloudflareResponse) error {
```
cblecker, line 121: Instead of mutating the input response inside each time, should we just return `(*CloudflareResponse, error) from this?

pkg/controller/certificaterequest/cloudflare.go
mbarnes: Good idea  :+1:

pkg/controller/certificaterequest/cloudflare.go
mbarnes: Maybe?  I'm worried about data being left from the previous iteration, but I'm not real familiar with Golang's JSON unmarshalling.  In particular, if we get an `Authority` item (negative cache response) on the first attempt, will that item still be in the Golang struct if the second JSON response has no `Authority` member?

The reallocation ensures we have a clean struct.

pkg/controller/certificaterequest/cloudflare.go
mbarnes: Since `VerifyDnsResourceRecordUpdate` carries the response struct into its next loop iteration to check for a negative cache result, I thought it would be _slightly_ clearer for `VerifyDnsResourceRecordUpdate` to own the struct.  (Maybe not relevant; still coming at it from a C mindset.)

mbarnes:
Force-pushed @cblecker's suggested changes for improved readability.

PR #138, opened 2020-05-02
OSD-3342 Created new tests for renewing certificate and certificate f…

luis-falcon:
Added some tests to bring the test coverage of the certman-operator above 20%

### DIFF
--- a/pkg/controller/certificaterequest/certificate_test.go
+++ b/pkg/controller/certificaterequest/certificate_test.go
@@ -17,7 +17,6 @@ limitations under the License.
 package certificaterequest
 
 import (
-	"crypto/x509"
 	"testing"
 
 	certmanv1alpha1 "github.com/openshift/certman-operator/pkg/apis/certman/v1alpha1"
@@ -26,22 +25,22 @@ import (
 
 func TestParseCertificateData(t *testing.T) {
 	tests := []struct {
-		name string
-		data []byte
-		want *x509.Certificate
+		name    string
+		data    []byte
+		wantNil bool
 	}{
 		{
-			name: "private key test",
-			data: leAccountPrivKey,
-			want: nil,
+			name:    "private key test",
+			data:    leAccountPrivKey,
+			wantNil: true,
 		},
 	}
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
 			got, _ := ParseCertificateData(test.data)
-			if got != test.want {
-				t.Errorf("ParseCertificateData() Error: Got = %v - Want = %v", got, test.want)
+			if (got == nil) != test.wantNil {
+				t.Errorf("ParseCertificateData() Error: Got = %v", got)
 			}
 		})
 	}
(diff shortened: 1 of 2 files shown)


pkg/controller/certificaterequest/certificate_test.go
```diff
@@ -0,0 +1,73 @@
...
+limitations under the License.
+*/
+
+package certificaterequest
+
+import (
+	"testing"
+
+	certmanv1alpha1 "github.com/openshift/certman-operator/pkg/apis/certman/v1alpha1"
+	//logrTesting "github.com/go-logr/logr/testing"
```
yithian, line 23: this may be just a personal pet peeve, but i'd like to see commented-out-code removed instead of committed

pkg/controller/certificaterequest/certificate_test.go
```diff
@@ -0,0 +1,73 @@
...
+			data:    leAccountPrivKey,
+			wantNil: true,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			got, _ := ParseCertificateData(test.data)
+			if (got == nil) != test.wantNil {
+				t.Errorf("ParseCertificateData() Error: Got = %v", got)
```
yithian, line 43: Why not test for an error being returned here instead? It seems more direct, unless I'm missing something, which is definitely possible.

```suggestion
		wantErr bool
	}{
		{
			name:    "private key test",
			data:    leAccountPrivKey,
			wantErr: true,
		},
	}

	for _, test := range tests {
		t.Run(test.name, func(t *testing.T) {
			_, err := ParseCertificateData(test.data)
			if (err != nil) && test.wantErr {
				t.Errorf("ParseCertificateData() Got unexpected error: %v", err)
```

pkg/controller/certificaterequest/certificate_test.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	var c = setUpEmptyTestClient(t)
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			got, _ := GetCertificate(c, test.testCert)
+
+			if (got == nil) != test.wantNil {
+				t.Errorf("GetCertificate() Error: Got = %v", got)
```
yithian, line 69: Same question here: why not just test for an error if that's what we're looking for?

```suggestion
		wantErr  bool
	}{
		{
			name:     "Test cert request with no secret",
			testCert: certRequest,
			wantErr:  true,
		},
	}

	var c = setUpEmptyTestClient(t)

	for _, test := range tests {
		t.Run(test.name, func(t *testing.T) {
			_, err := GetCertificate(c, test.testCert)

			if (err != nil) && test.wantErr {
				t.Errorf("GetCertificate() Got unexpected error: %v", err)
```

pkg/controller/certificaterequest/certificate_test.go
luis-falcon: Good catch. Thanks

pkg/controller/certificaterequest/certificate_test.go
luis-falcon: Implemented changes.

PR #142, opened 2020-05-11
prepare the repo for codecov.io representation

yithian:


cblecker:
Two things I missed to point out that we should follow up on:
- We should ensure `*.tmp` files are added to the gitignore (`*.out` is already in there)
- `TESTTARGETS` in the standard.mk isn't used anymore and can probably be cleaned up

PR #146, opened 2020-05-25
Add support for Azure DNS Zone

hamza3202:
This adds support for Azure DNS Zone.


pkg/clients/azure/dns.go
```diff
@@ -0,0 +1,236 @@
...
+		if err != nil {
+			reqLogger.Error(err, "Error while deleting Write Access record")
+			return false, err
+		}
+		// If Write and Delete are successfull return clean.
+		return true, nil
+	}
+
+	reqLogger.Error(err, "Private DNS zone is not allowed")
+	return false, nil
```
yithian, line 163: Personally, I think it makes more sense with this logic reversed. Is this something you considered?

```suggestion
	if zone.ZoneType != "Private" {
		reqLogger.Error(err, "Private DNS zone is not allowed")
		return false, nil
	}
	
	// Build the test record
	_, err := c.createTxtRecord(reqLogger, recordKey, "\"txt_entry\"", *zone.Name)

	if err != nil {
		return false, err
	}

	// After successfull write test clean up the test record and test deletion of that record.
	_, err = c.recordSetsClient.Delete(context.TODO(), c.resourceGroupName, *zone.Name, recordKey, dns.TXT, "")

	if err != nil {
		reqLogger.Error(err, "Error while deleting Write Access record")
		return false, err
	}
	// If Write and Delete are successfull return clean.
	return true, nil
```

pkg/clients/azure/dns_test.go
```diff
@@ -0,0 +1,239 @@
...
+
+	t.Run("returns a client if the credentials is set", func(t *testing.T) {
+		testClient := setUpTestClient(t, getAzureSecret(validSecretData))
+
+		_, err := NewClient(testClient, testHiveAzureSecretName, testHiveNamespace, testHiveResourceGroupName)
+
+		if err != nil {
+			t.Errorf("unexpected error when creating the client: %q", err)
+		}
+	})
```
yithian, line 51: these should be set up as table tests, like `TestGetAzureCredentialsFromSecret`, below

pkg/clients/azure/dns.go
hamza3202: Agreed!

pkg/clients/azure/dns_test.go
hamza3202: Done!

hamza3202:
@yithian I have handled the feedback. Please let me know if you need anything more from my side.

yithian:
seems sane

/lgtm

pkg/clients/azure/dns.go
```diff
@@ -0,0 +1,236 @@
...
+
+	recordKey := "_certman_access_test." + *zone.Name
+
+	if zone.ZoneType != "Private" {
+		reqLogger.Error(err, "Private DNS zone is not allowed")
+		return false, nil
+	}
+	
+	// Build the test record
+	_, err := c.createTxtRecord(reqLogger, recordKey, "\"txt_entry\"", *zone.Name)
```
yithian, line 149: since err was used above, the `:=` operator causes an error here: `./dns.go:149:9: no new variables on left side of :=`

```suggestion
	_, err = c.createTxtRecord(reqLogger, recordKey, "\"txt_entry\"", *zone.Name)
```

hamza3202:
@yithian the tests pass now with the new changes. Can you please re-run the pipeline?

pkg/clients/azure/dns.go
```diff
@@ -0,0 +1,236 @@
...
+		return false, err
+	}
+
+	recordKey := "_certman_access_test." + *zone.Name
+
+	if zone.ZoneType != "Private" {
+		reqLogger.Error(err, "Private DNS zone is not allowed")
+		return false, nil
+	}
+
```
yithian, line 147: gofmt is failing on trailing whitespace here

```suggestion

```

hamza3202: Remove trailing whitespaces

PR #148, opened 2020-06-03
Fix Dns zone type check

hamza3202:
This PR fixes a bug in azure support where it only allows using private hosted zone whereas it should actually allow only public hosted zones.

The bug was introduced in https://github.com/openshift/certman-operator/pull/146


hamza3202:
@yithian  @c-e-brumm this is a very small but critical fix. Can we have this merged ASAP?

yithian:
I want to stress the importance of unit testing to catch problems like this before they get merged into the code base.

/lgtm

PR #151, opened 2020-07-31
Remove duplicate CRD definition

cblecker:
Looks like two versions of the CRD were created when the operator-sdk was updated.

Thanks to @2uasimojo for noticing this.


2uasimojo:
/lgtm
but I'm following with an update to the README that includes fixing a reference to this file.

2uasimojo:
> but I'm following with an update to the README that includes fixing a reference to this file.

#152

PR #154, opened 2020-08-12
Optimize CertificateRequest Reconcile Behavior

dustman9000:
These changes fix the reconcile behavior of the CertificateRequest reconcile to requeue after a certain duration such that subsequent CRs can be processed. The old behavior was effectively processing each CR in serial and this changes it to a rapid succession.

### DIFF
--- a/hack/test/scale-test.sh
+++ b/hack/test/scale-test.sh
@@ -3,17 +3,22 @@
 # 1. Spoof a ClusterDeployment to replicate running in Hive,
 #    which will trigger certificates to be generated.
 set -o errexit
-set -o errtrace
-set -o nounset
 set -o pipefail
 
-BASE_DOMAIN=fidata.io
-
-TOTAL_COUNT=${1:-10}
+BASE_DOMAIN=$1
+TOTAL_COUNT=$2
 
 NAMES_ARR=()
 START_TIMES_ARR=()
 DURATION_TIMES_ARR=()
+
+function usage() {
+  echo "Usage:"
+  echo "    $0 <BASE_DOMAIN> <COUNT>"
+  echo
+  echo "Example:"
+  echo "    $0 acme.io 15"
+}
 
 function cleanup() {
     for i in $(seq 1 $TOTAL_COUNT)
@@ -23,13 +28,23 @@ function cleanup() {
     done
 }
 
+if [ -z "${BASE_DOMAIN}" ]; then
+    usage
+    exit 1
+fi
+
+if [ -z "${TOTAL_COUNT}" ]; then
+    usage
+    exit 1
+fi
+
 trap cleanup INT TERM EXIT
 
 for i in $(seq 1 $TOTAL_COUNT)
 do
     CD_START_TIME=`date +%s`
     RAND0=$(head /dev/urandom | tr -dc a-z | head -c 4 ; echo '')
-    RAND2=$(head /dev/urandom | tr -dc A-Za-z0-9 | head -c 8 ; echo '')
+    RAND1=$(head /dev/urandom | tr -dc A-Za-z0-9 | head -c 8 ; echo '')
 
     NAMES_ARR[$i]=${RAND0}
     START_TIMES_ARR[$i]=${CD_START_TIME}
@@ -62,8 +77,8 @@ spec:
       name: ${RAND0}-admin-kubeconfig
     adminPasswordSecretRef:
       name: ${RAND0}-admin-password
-    clusterID: ${RAND2}
-    infraID: ${RAND2}
+    clusterID: ${RAND1}
+    infraID: ${RAND1}
   clusterName: ${RAND0}
   ingress:
   - domain: ${RAND0}.${BASE_DOMAIN}
--- a/hack/test/scale-test.sh
+++ b/hack/test/scale-test.sh
@@ -1,0 +1,117 @@
+#!/bin/bash
+# This script will test certman-operator in minikube by doing the following:
+# 1. Spoof a ClusterDeployment to replicate running in Hive,
+#    which will trigger certificates to be generated.
+set -o errexit
+set -o errtrace
+set -o nounset
+set -o pipefail
+
+BASE_DOMAIN=fidata.io
+
+TOTAL_COUNT=${1:-10}
+
+NAMES_ARR=()
+START_TIMES_ARR=()
+DURATION_TIMES_ARR=()
+
+function cleanup() {
+    for i in $(seq 1 $TOTAL_COUNT)
+    do
+        NAME=${NAMES_ARR[$i]}
+        kubectl delete clusterdeployment ${NAME}
+    done
+}
+
+trap cleanup INT TERM EXIT
+
+for i in $(seq 1 $TOTAL_COUNT)
+do
+    CD_START_TIME=`date +%s`
+    RAND0=$(head /dev/urandom | tr -dc a-z | head -c 4 ; echo '')
+    RAND2=$(head /dev/urandom | tr -dc A-Za-z0-9 | head -c 8 ; echo '')
+
+    NAMES_ARR[$i]=${RAND0}
+    START_TIMES_ARR[$i]=${CD_START_TIME}
+
+    cat <<EOF | kubectl apply -f -
+apiVersion: hive.openshift.io/v1
+kind: ClusterDeployment
+metadata:
+  creationTimestamp: null
+  labels:
+    api.openshift.com/ccs: "false"
+    api.openshift.com/environment: staging
+    api.openshift.com/managed: "true"
+    api.openshift.com/name: ${RAND0}
+    api.openshift.com/service-lb-quota: "0"
+    api.openshift.com/storage-quota-gib: "100"
+    hive.openshift.io/cluster-platform: aws
+    hive.openshift.io/cluster-region: us-west-2
+    hive.openshift.io/cluster-type: managed
+  name: ${RAND0}
+spec:
+  baseDomain: ${BASE_DOMAIN}
+  certificateBundles:
+  - certificateSecretRef:
+      name: ${RAND0}-primary-cert-bundle-secret
+    generate: true
+    name: ${RAND0}-primary-cert-bundle
+  clusterMetadata:
+    adminKubeconfigSecretRef:
+      name: ${RAND0}-admin-kubeconfig
+    adminPasswordSecretRef:
+      name: ${RAND0}-admin-password
+    clusterID: ${RAND2}
+    infraID: ${RAND2}
+  clusterName: ${RAND0}
+  ingress:
+  - domain: ${RAND0}.${BASE_DOMAIN}
+    name: default
+    servingCertificate: ${RAND0}-primary-cert-bundle
+  installed: true
+  manageDNS: true
+  platform:
+    aws:
+      credentialsSecretRef:
+        name: aws
+      region: us-west-2
+  provisioning:
+    imageSetRef:
+      name: openshift-v4.3.25
+    installConfigSecretRef:
+      name: install-config
+    sshPrivateKeySecretRef:
+      name: ssh
+  pullSecretRef:
+    name: pull
+EOF
+done
+
+ISSUED_COUNT=0
+
+echo "Waiting for $TOTAL_COUNT certificates to be issued..."
+
+while [ $ISSUED_COUNT -lt $TOTAL_COUNT ]
+do
+    for i in $(seq 1 $TOTAL_COUNT)
+    do
+        NAME=${NAMES_ARR[$i]}
+
+        if [ ! ${DURATION_TIMES_ARR[$i]+abc} ]; then
+            #echo "Checking if ${NAME}.${BASE_DOMAIN} certificate was issued"
+            ISSUED=$(kubectl get certificaterequests ${NAME}-${NAME}-primary-cert-bundle -o json | jq .status.issued)
+
+            if [ "${ISSUED}" == "true" ]; then
+                CERT_ISSUED_END_TIME=`date +%s`
+                CERT_ISSUE_DURATION=$((CERT_ISSUED_END_TIME-CD_START_TIME))
+                DURATION_TIMES_ARR[$i]=${CERT_ISSUE_DURATION}
+                echo "${NAME} ISSUE DURATION: ${CERT_ISSUE_DURATION} seconds"
+                ((ISSUED_COUNT++))
+            fi
+        fi
+    done
+    sleep 1
+done
+
+cleanup
--- a/pkg/controller/certificaterequest/certificaterequest_controller.go
+++ b/pkg/controller/certificaterequest/certificaterequest_controller.go
@@ -19,6 +19,7 @@ package certificaterequest
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -30,9 +31,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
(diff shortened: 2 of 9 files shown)


dustman9000:
> Please add unit test coverage for the code being added.

All done. Added unit tests around utils and cloudflare functions.

cblecker:
Thanks for this @dustman9000! This is quite a big change, so I'd like to get this reviewed throughly.

/hold
Holding for reviews.

pkg/controller/certificaterequest/utils_test.go
```diff
@@ -0,0 +1,124 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			r := &ReconcileCertificateRequest{
+				client:        tt.fields.client,
+				scheme:        tt.fields.scheme,
+				clientBuilder: tt.fields.clientBuilder,
+			}
+			if err := r.IncrementDNSVerifyAttempts(tt.args.reqLogger, tt.args.cr); (err != nil) != tt.wantErr {
+				t.Errorf("ReconcileCertificateRequest.IncrementDNSVerifyAttempts() error = %v, wantErr %v", err, tt.wantErr)
+			}
```
yithian, line 76: It looks like this just tests that `IncrementDNSVerifyAttemps()` doesn't throw an error and doesn't test if the function actually increments the attempts counter. Is that accurate?

hack/test/scale-test.sh
```diff
@@ -0,0 +1,117 @@
+#!/bin/bash
+# This script will test certman-operator in minikube by doing the following:
+# 1. Spoof a ClusterDeployment to replicate running in Hive,
+#    which will trigger certificates to be generated.
+set -o errexit
+set -o errtrace
+set -o nounset
+set -o pipefail
+
+BASE_DOMAIN=fidata.io
```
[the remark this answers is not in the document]
dustman9000: RH does not own this domain, this is my personal domain managed in my personal AWS account. I'll fix this var to be explicitly generic and error out if unset.

pkg/controller/certificaterequest/cloudflare.go
```diff
@@ -52,82 +52,74 @@ type CloudflareResponse struct {
...
+
+	// Success if returned record value equals expetect txt value
+	if strings.EqualFold(cfName, fqdn) && cfData == txtValue {
+		reqLogger.Info(fmt.Sprintf("record %v has been updated to value %v.", fqdn, txtValue))
+		return 0
 	}
 
-	errMsg := fmt.Sprintf("unable to verify that resource record %v has been updated to value %v after %v attempts.", fqdn, txtValue, maxAttemptsForDnsPropagationCheck)
+	errMsg := fmt.Sprintf("unable to verify that resource record %v has been updated to value %v.", fqdn, txtValue)
 	reqLogger.Error(errors.New(errMsg), errMsg)
```
jharrington22, line 112: I see that this is carried over but it doesn't feel like we should be returning an error here if we're waiting on caching or the record hasn't propagated yet.

pkg/controller/certificaterequest/utils.go
```diff
@@ -50,3 +55,44 @@ func GetSecret(kubeClient client.Client, secretName, namespace string) (*corev1.
 
 	return s, nil
 }
+
+// GetDNSVerifyAttempts returns the attempts based on the annotation
+func GetDNSVerifyAttempts(cr *certmanv1alpha1.CertificateRequest) int {
+	ann := certmanv1alpha1.CertmanOperatorFinalizerLabel + "/" + dnsCheckAttemptsLabel
```
jharrington22, line 61: Should move this to a const

pkg/controller/certificaterequest/issue_certificate.go
```diff
@@ -81,64 +81,75 @@ func (r *ReconcileCertificateRequest) IssueCertificate(reqLogger logr.Logger, cr
...
+
+		// If this is the first DNS check, then back off for 60s to avoid negative cache TTL
+		if verifyAttempts == 0 {
+			return initialWaitPeriodDNSPropagationCheck, nil
 		}
 
-		dnsChangesVerified := VerifyDnsResourceRecordUpdate(reqLogger, fqdn, DNS01KeyAuthorization)
-		if !dnsChangesVerified {
-			return fmt.Errorf("cannot complete Let's Encrypt challenege as DNS changes could not be verified")
+		sleepDuration := VerifyDnsResourceRecordUpdate(reqLogger, fqdn, DNS01KeyAuthorization)
```
jharrington22, line 128: `waitPeriod` to be consistent?

pkg/controller/certificaterequest/utils.go
```diff
@@ -50,3 +55,44 @@ func GetSecret(kubeClient client.Client, secretName, namespace string) (*corev1.
...
+
+	// Make sure we have the latest version before we update
+	err := r.client.Get(context.TODO(), types.NamespacedName{
+		Namespace: cr.Namespace,
+		Name:      cr.Name,
+	}, cr)
+
+	if metav1.HasAnnotation(cr.ObjectMeta, ann) {
+		i, err := strconv.Atoi(cr.Annotations[ann])
+		if err == nil && i == attempts {
```
jharrington22, line 86: What if there is an error?

pkg/controller/certificaterequest/issue_certificate.go
```diff
@@ -36,40 +36,40 @@ import (
 // IssueCertificate validates DNS write access then assess letsencrypt endpoint (prod or stage) based on leclient url.
 // It then iterates through the CertificateRequest.Spec.DnsNames, authorizes to letsencrypt and sets a challenge in the
 // form of resource record. Certificates are then generated and issued to kubernetes via corev1.
-func (r *ReconcileCertificateRequest) IssueCertificate(reqLogger logr.Logger, cr *certmanv1alpha1.CertificateRequest, certificateSecret *corev1.Secret) error {
+func (r *ReconcileCertificateRequest) IssueCertificate(reqLogger logr.Logger, cr *certmanv1alpha1.CertificateRequest, certificateSecret *corev1.Secret) (waitPeriod int, err error) {
```
jharrington22, line 39: Consider returning a pointer for `waitPeriod` so we're checking `nil` and not `-1`. I'm not entirely sure if that will leak mem though.

hack/test/scale-test.sh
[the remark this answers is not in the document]
lisa: Will Let's Encrypt support using one of the RFC-designated testing domains?

pkg/controller/certificaterequest/cloudflare.go
```diff
@@ -52,82 +52,74 @@ type CloudflareResponse struct {
...
+
+	// If there is no answer field, this is likely an expected NXDOMAIN response.
+	if len(response.Answers) == 0 {
+		reqLogger.Info("no answers received from cloudflare; likely not propagated")
+		return waitPeriod
+	}
+
+	// Trim any trailing dot from the answer name and quotes from the data.
+	cfName := strings.TrimSuffix(response.Answers[0].Name, ".")
+	cfData := strings.Trim(response.Answers[0].Data, "\"")
```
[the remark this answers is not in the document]
dustman9000: I'm not sure adding a function for these 2 lines, really adds much.

pkg/controller/certificaterequest/cloudflare.go
dustman9000: Updated to an Info log.

pkg/controller/certificaterequest/issue_certificate.go
dustman9000: Yeah, I think returning multiple values is ok here. I'm a little hesitant to add something that would leak memory here.

pkg/controller/certificaterequest/issue_certificate.go
dustman9000: Done.

pkg/controller/certificaterequest/utils.go
dustman9000: Done.

pkg/controller/certificaterequest/utils_test.go
dustman9000: Thanks. I've updated the test to actually test the value as well.

pkg/controller/certificaterequest/utils.go
dustman9000: Fixed.

pkg/controller/certificaterequest/cloudflare.go
```diff
@@ -52,82 +52,74 @@ type CloudflareResponse struct {
...
-
-		if strings.EqualFold(cfName, fqdn) && cfData == txtValue {
-			return true
-		}
-
-		reqLogger.Info("could not validate DNS propagation for " + fqdn)
+	// Set default wait period to 30s
+	waitPeriod := defaultWaitPeriodDNSPropagationCheck
+
+	reqLogger.Info(fmt.Sprintf("attempting to verify resource record %v has been updated with value %v", fqdn, txtValue))
```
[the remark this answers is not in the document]
dustman9000: I think this log message adds value and I'm actually removing logging from this function believe it or not.

pkg/controller/certificaterequest/certificaterequest_controller.go
```diff
@@ -243,15 +270,22 @@ func (r *ReconcileCertificateRequest) createCertificateSecret(reqLogger logr.Log
...
 			reqLogger.Error(updateErr, updateErr.Error())
 		}
 		reqLogger.Error(err, err.Error())
 		return reconcile.Result{}, err
 	}
+	if waitPeriod > 0 {
+		if err := r.IncrementDNSVerifyAttempts(reqLogger, cr); err != nil {
+			reqLogger.Error(err, "Error Incrementing DNS verify check attempts")
+		}
+		reqLogger.Info(fmt.Sprintf("dns validation failed, requeue reconcile after %d seconds", waitPeriod))
```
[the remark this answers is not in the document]
dustman9000: I think this Info log adds value here as this is the only log indicating that the waitPeriod was greater than zero. Even though it's not an error, it adds context.

hack/test/scale-test.sh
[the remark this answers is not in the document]
dustman9000: Not sure I know what you're asking. This domain is intended to match up with the aws secret in the cluster.

pkg/controller/certificaterequest/certificaterequest_controller.go
```diff
@@ -168,9 +187,17 @@ func (r *ReconcileCertificateRequest) Reconcile(request reconcile.Request) (reco
 	}
 
 	if shouldReissue {
-		err := r.IssueCertificate(reqLogger, cr, found)
+		waitPeriod, err := r.IssueCertificate(reqLogger, cr, found)
 		if err != nil {
 			return reconcile.Result{}, err
+		} else if waitPeriod == -1 {
+			return reconcile.Result{}, fmt.Errorf("Error issuing certificate")
+		}
```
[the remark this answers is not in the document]
dustman9000: It's true that waitPeriod and err are couple together on error. There was a situation that err was nil, but waitPeriod was non-zero. I added the extra check to be certain and to future proof things.

pkg/controller/certificaterequest/certificaterequest_controller.go
```diff
@@ -243,15 +270,22 @@ func (r *ReconcileCertificateRequest) createCertificateSecret(reqLogger logr.Log
 		return reconcile.Result{}, err
 	}
 
-	err := r.IssueCertificate(reqLogger, cr, certificateSecret)
-	if err != nil {
+	waitPeriod, err := r.IssueCertificate(reqLogger, cr, certificateSecret)
+	if err != nil || waitPeriod == -1 {
```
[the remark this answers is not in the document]
dustman9000: There was a situation where err could be nil, and waitPeriod was non-zero, so I think the check does add value here.

pkg/controller/certificaterequest/cloudflare.go
```diff
@@ -52,82 +52,74 @@ type CloudflareResponse struct {
 	Authority []CloudflareAnswer   `json:"Authority"`
 }
 
-// VerifyDnsResourceRecordUpdate verifies the presence of a TXT record with Cloudflare DNS.
-func VerifyDnsResourceRecordUpdate(reqLogger logr.Logger, fqdn string, txtValue string) bool {
+// VerifyDnsResourceRecordUpdate verifies the presence of a TXT record with Cloudflare DNS. Returns 0 on success.
+func VerifyDnsResourceRecordUpdate(reqLogger logr.Logger, fqdn string, txtValue string) int {
```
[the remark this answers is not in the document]
dustman9000: I thought about that, but there was only one situation where there could be an error. I felt it unnecessary to propagate that single error to the caller. I feel the error message is enough here.

hack/test/scale-test.sh
```diff
@@ -0,0 +1,117 @@
...
+    done
+}
+
+trap cleanup INT TERM EXIT
+
+for i in $(seq 1 $TOTAL_COUNT)
+do
+    CD_START_TIME=`date +%s`
+    RAND0=$(head /dev/urandom | tr -dc a-z | head -c 4 ; echo '')
+    RAND2=$(head /dev/urandom | tr -dc A-Za-z0-9 | head -c 8 ; echo '')
```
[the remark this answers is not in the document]
dustman9000: Fixed :)

pkg/controller/certificaterequest/issue_certificate_test.go
```diff
@@ -32,10 +32,13 @@ func TestIssueCertificate(t *testing.T) {
...
 
-		err := rcr.IssueCertificate(nullLogger, certRequest, certSecret)
+		waitPeriod, err := rcr.IssueCertificate(nullLogger, certRequest, certSecret)
 
 		if err == nil {
 			t.Error("expected an error")
 		}
+		if waitPeriod != -1 {
+			t.Error("expected a wait period of -1")
+		}
```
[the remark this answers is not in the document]
dustman9000: There was a situation where err could be nil, and waitPeriod was non-zero, so I think the check does add value here.

hack/test/scale-test.sh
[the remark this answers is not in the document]
lisa: @dustman9000 [RFC 6761](https://tools.ietf.org/html/rfc6761) talks about a variety of reserved domain names, such as example.com, and others.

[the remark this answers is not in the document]
dustman9000: Ah, gotcha. No, Let's Encrypt doesn't support test/reserved domains as the Certificate Authority needs a publicly resolvable domain to issue a certificate. For testing, we can use LE's staging servers for testing API calls.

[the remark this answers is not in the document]
lisa: @dustman9000 Gotcha! 👍  Was a passing curiosity :)

pkg/controller/certificaterequest/issue_certificate.go
jharrington22: Would like @cblecker opinion here

pkg/controller/certificaterequest/cloudflare.go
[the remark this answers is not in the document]
jharrington22: I was thinking that it would help support unit tests

pkg/controller/certificaterequest/cloudflare.go
```diff
@@ -52,82 +51,73 @@ type CloudflareResponse struct {
...
+	// If the SOA TTL exceeds this, give up immediately.
+	if negativeCacheTTL > maxNegativeCacheTTL {
+		reqLogger.Info("negative cache TTL is too long; giving up")
+		return maxNegativeCacheTTL
+	}
+
+	// If the TTL is shorter than the default wait time, disregard.
+	if negativeCacheTTL > waitPeriod {
+		waitPeriod = negativeCacheTTL
+	}
```
cblecker, line 92: This doesn't exactly do what we want it to after the rewrite, because this is happening *after* the fetch. `negativeCacheTTL` is set to anything but 0, then there had to have been a negative cache response meaning we need to sleep before we retry.

pkg/controller/certificaterequest/certificaterequest_controller.go
```diff
@@ -71,9 +74,24 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	if err != nil {
 		return err
 	}
+	// Only filter on create/delete/update events
```
cblecker, line 77: What was the generic event that was triggering this before?

dustman9000:
Yes, that's a much more simple change than this PR. I'm ok with closing this in favor of 156.

pkg/controller/certificaterequest/cloudflare.go
jharrington22: @cblecker I don't understand what the issue is here? the end result could be that we put this request aside for `netgativeCacheTTL` and try again.

dustman9000:
Closing in favor of #156

PR #156, opened 2020-08-20
Set MaxConcurrentReconciles for CertificateRequest CR to 5

cblecker:


pkg/controller/certificaterequest/certificaterequest_controller.go
```diff
@@ -67,7 +67,12 @@ func newReconciler(mgr manager.Manager) reconcile.Reconciler {
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
 func add(mgr manager.Manager, r reconcile.Reconciler) error {
 
-	c, err := controller.New("certificaterequest-controller", mgr, controller.Options{Reconciler: r})
+	o := controller.Options{
+		Reconciler:              r,
+		MaxConcurrentReconciles: 5,
```
dustman9000, line 72: How did you land on 5? Should this turn into a const?

gregsheremeta: hive is making this configurable in a CR [https://github.com/openshift/hive/pull/1083]

suggest you also make it configurable somehow

cblecker: @gregsheremeta That's a cool idea, but would blow the scope of this for today. We can look at this in a future iteration.

For now, I've made it a const at the top of the file

gregsheremeta: yeah agreed. Totally a suggestion for the far future.

PR #159, opened 2020-09-15
Update `generate` target for operator-sdk 0.16.0

2uasimojo:
This repository is using operator-sdk 0.16.0, in which `generate openapi` is deprecated. Replace it with `generate crds`.

The file generated by `generate openapi` is unused, so it is removed.


cblecker:
@2uasimojo is this still needed with the eventual move to boilerplate?

2uasimojo:
> @2uasimojo is this still needed with the eventual move to boilerplate?

This PR can probably be closed. But the more important question remains: Do we need openapi? Right now I've been hacking an override into repositories that were generating openapi (e.g. [mvo](https://github.com/openshift/managed-velero-operator/blob/4abf891f1005b40bd9697a0c51fd2c63ea0649d0/Makefile#L7-L25) because I don't have that answer yet.

2uasimojo:
For those of you following along at home, we added openapi generation in boilerplate via https://github.com/openshift/boilerplate/pull/111. Consumers at or above this level are forced to check in generated openapi and keep it up to date.

PR #161, opened 2020-09-30
Adding new counter for CertificateRequest

bdematte:


bdematte:
tested locally : 
```
# HELP certman_operator_certificate_requests_count Report the current count of Certificate Requests
# TYPE certman_operator_certificate_requests_count gauge
certman_operator_certificate_requests_count{name="certman-operator"} 1
```

pkg/controller/certificaterequest/certificaterequest_controller.go
```diff
@@ -141,6 +144,7 @@ func (r *ReconcileCertificateRequest) Reconcile(request reconcile.Request) (reco
 	// Add finalizer if not exists
 	if !utils.ContainsString(cr.ObjectMeta.Finalizers, certmanv1alpha1.CertmanOperatorFinalizerLabel) {
 		reqLogger.Info("adding finalizer to the certificate request")
+		localmetrics.IncrementCertRequestsCounter()
```
arjunrn, line 147: If the operator pod is restarted and there are some existing `CertfificateRequest` objects then they will get counted twice. Once during the `CheckInitCounter` and then while adding the finalizer.

pkg/localmetrics/localmetrics.go
```diff
@@ -56,6 +60,10 @@ var (
 		Help:        "The duration it takes to reconcile a ClusterDeployment",
 		ConstLabels: prometheus.Labels{"name": "certman-operator"},
 	})
+	MetricCertRequestsCount = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
arjunrn, line 63: Does this have to be a `Vec` type?

pkg/controller/certificaterequest/certificaterequest_controller.go
bdematte: Yes, that's why I put the limitation in the describtion of the Init method 
``` 
 Current version does not support well multiple instances of the operator to run on the same Hive cluster 
```
We discussed it with James and thought it would be OK for a start.

arjunrn: This will affect the values returned even when there is a single pod. say there are bunch of certificate requests which are created. the operator starts dealing with them but crashes without adding finalizers. when it restarts it counts all of them and then when it adds the finalizers it counts them again.

pkg/localmetrics/localmetrics.go
bdematte: The only interest (on top of the alignment with other gauges) is to ease future addition of labels, but it does not need to. I can update it with ConstLabel for the operator name indeed. 
What do we prefer ?

pkg/controller/certificaterequest/certificaterequest_controller.go
bdematte: ah yes, thanks.. Will filter to count only the ones having the finalzer

pkg/localmetrics/localmetrics.go
arjunrn: Let's keep it a simple gauge for now.

PR #163, opened 2020-10-14
Add certificate issuance rate

bdematte:
This PR is adding a counter on the number of issued certificate.
In addition, it is adding missing error in 1 scenario detected while testing the code (in case we can't get write access to DNS record, we were forwarding a nil error instead of an actual one, leading secret to be created while no certificate).

This PR is based on code from PR161 - Adding new counter for CertificateRequest (as touching the same pieces of code) and should only be merged after this one to have a clear change history so setting this PR as WIP for now.

### DIFF
--- a/pkg/controller/certificaterequest/issue_certificate.go
+++ b/pkg/controller/certificaterequest/issue_certificate.go
@@ -22,6 +22,7 @@ import (
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/pem"
+	"errors"
 	"fmt"
 
 	"github.com/go-logr/logr"
@@ -57,6 +58,7 @@ func (r *ReconcileCertificateRequest) IssueCertificate(reqLogger logr.Logger, cr
 	if proceed {
 		reqLogger.Info("write permissions for DNS has been validated")
 	} else {
+		err = errors.New("failed to get write access to DNS record")
 		reqLogger.Error(err, "failed to get write access to DNS record")
 		return err
 	}
--- a/pkg/localmetrics/localmetrics.go
+++ b/pkg/localmetrics/localmetrics.go
@@ -66,6 +66,10 @@ var (
 		Help: "Report the current count of Certificate Requests",
 		ConstLabels: prometheus.Labels{"name": "certman-operator"},
 	})
+	MetricCertIssuanceRate = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "certman_operator_issued_certficates_count",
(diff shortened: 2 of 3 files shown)


pkg/localmetrics/localmetrics.go
```diff
@@ -56,6 +60,14 @@ var (
 		Help:        "The duration it takes to reconcile a ClusterDeployment",
 		ConstLabels: prometheus.Labels{"name": "certman-operator"},
 	})
+	MetricCertRequestsCount = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "certman_operator_certificate_requests_count",
+		Help: "Report the current count of Certificate Requests",
+	}, []string{"name"})
+	MetricCertIssuanceRate = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "certman_operator_certificate_issuance_rate",
```
arjunrn, line 68: This looks more like a counter than a gauge because the value only goes up. Also rate shouldn't be part of the name because the rate is calculated by the query later.

bdematte: Agreed on the type. Regarding the `rate` in the name of the metric, it is because it should not be used as a counter (as never initialized, it does not represent an actual number of certificate). 
`certman_operator_issued_certificates_counter` would be better ?

arjunrn: It doesn't need to be initialized because it will be zero by when created. As for the name `certman_operator_issued_certficates_count` would be in line with naming conventions used in most places.

bdematte:
/hold
Pending merge of https://github.com/openshift/certman-operator/pull/161

bdematte:
rebased after PR#161 merge
/unhold

pkg/controller/certificaterequest/issue_certificate.go
```diff
@@ -57,6 +58,7 @@ func (r *ReconcileCertificateRequest) IssueCertificate(reqLogger logr.Logger, cr
 	if proceed {
 		reqLogger.Info("write permissions for DNS has been validated")
 	} else {
+		err = errors.New("failed to get write access to DNS record")
```
yithian, line 61: why not use the err returned by `dnsClient.ValidateDNSWriteAccess(reqLogger, cr)` ?

bdematte: In that case, the error is nil per the previous if loop, however, we logically handled it as an error hence creating a new error to align the behaviour

PR #165, opened 2020-10-20
Update yq syntax and fix saas file location

jharrington22:
The image used for yq was change to be the golang version which has a completely different syntax to the original python version. The entry point for the container is also not correct hence needing the extra yq.

The path to discover what version of the operator promoted to production has changed in app-interface, this has been updated.

### DIFF
--- a/hack/app_sre_create_image_catalog.sh
+++ b/hack/app_sre_create_image_catalog.sh
@@ -28,7 +28,7 @@ REMOVED_VERSIONS=""
 if [[ "$REMOVE_UNDEPLOYED" == true ]]; then
     DEPLOYED_HASH=$(
         curl -s "https://gitlab.cee.redhat.com/service/app-interface/raw/master/data/services/osd-operators/cicd/saas/saas-${_OPERATOR_NAME}.yaml" | \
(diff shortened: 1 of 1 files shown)


hack/app_sre_create_image_catalog.sh
```diff
@@ -28,7 +28,7 @@ REMOVED_VERSIONS=""
 if [[ "$REMOVE_UNDEPLOYED" == true ]]; then
     DEPLOYED_HASH=$(
         curl -s "https://gitlab.cee.redhat.com/service/app-interface/raw/master/data/services/osd-operators/cicd/saas/saas-${_OPERATOR_NAME}.yaml" | \
-        docker run --rm -i quay.io/app-sre/yq -r '.resourceTemplates[]|select(.name="${_OPERATOR_NAME}").targets[]|select(.namespace["$ref"]=="/services/osd-operators/namespaces/${_OPERATOR_NAME}-production.yml")|.ref'
+            docker run --rm -i quay.io/app-sre/yq yq r - "resourceTemplates[*].targets(namespace.\$ref==/services/osd-operators/namespaces/hivep01ue1/${_OPERATOR_NAME}.yml).ref"
```
yithian, line 31: do we need to be concerned about any other hive clusters? or will this always be `hivep01ue1`?

jharrington22: It wont always be `hivep01ue1` but this is at least a V4 cluster that will be around for a while.

PR #167, opened 2020-12-02
Create IsCertificateIssuerLE function to determine if certificate is LE

cblecker:
Create a more robust function, with testing, to determine if a certificate is issued by Let's Encrypt.

### DIFF
--- a/pkg/leclient/utils.go
+++ b/pkg/leclient/utils.go
@@ -18,6 +18,7 @@ package leclient
 
 import (
 	"context"
+	"crypto/x509/pkix"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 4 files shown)


pkg/leclient/utils.go
```diff
@@ -36,3 +37,20 @@ func GetSecret(kubeClient client.Client, secretName, namespace string) (*corev1.
 
 	return s, nil
 }
+
+// IsCertificateIssuerLE takes an issuer name on a certificate and determines if it's a Let's Encrypt CA
+func IsCertificateIssuerLE(issuer pkix.Name) bool {
+	if len(issuer.Organization) > 0 {
+		for _, o := range issuer.Organization {
+			if o == "Let's Encrypt" {
```
dustman9000, line 45: Can we make these constants?

pkg/leclient/utils.go
```diff
@@ -36,3 +37,20 @@ func GetSecret(kubeClient client.Client, secretName, namespace string) (*corev1.
...
+				return true
+			}
+		}
+	}
+
+	if issuer.CommonName == "Fake LE Intermediate X1" {
```
dustman9000, line 51: Suggest a constant for this string as well...

dustman9000:
Looks good! Just some minor comments on turning those strings into constants and re-using in unit-tests.

pkg/leclient/utils.go
cblecker: If we aren't reusing them, what's the benefit of a constant?

cblecker: I want the tests to be hard coded, as I don't want them touched if we mess with this function (the actual certs won't ever change). The tests also have more data then we actually check in this function.

So I'm intentionally *not* reusing strings with the tests.. and if not there, then we don't really reuse these strings anywhere (so the benefit of allocating a constant isn't there).

dustman9000: Fair enough, thanks.

PR #170, opened 2020-12-15
Onboard Boilerplate

dustman9000:
Changes necessary to bring certman-operator onboarded with the boilerplate framework.


Makefile
```diff
@@ -1,19 +1,11 @@
...
+.PHONY: boilerplate-update
+boilerplate-update:
+	@boilerplate/update
 
-# Push the docker image
-.PHONY: docker-push
-docker-push: push
+# TODO: Remove below targets once prow config standardized
+.PHONY: gobuild
+gobuild: go-build
```
2uasimojo, line 11: Don't need to do this, since prow isn't actually using `gobuild` directly. Instead, fix the [callout in the Dockerfile](https://github.com/openshift/certman-operator/pull/170/files?file-filters%5B%5D=.go&file-filters%5B%5D=.mk&file-filters%5B%5D=No+extension#diff-5aecd73bbcb9b7410938673269f0d4e676dd18509277b0ff2f2787c9d3e22604R8).

2uasimojo (commented):
This is legit, thanks @dustman9000. Just one thing that can be cleaned up.

2uasimojo:
Oh, one more thing: `hack/codecov.sh` can be removed. (`app_sre_pr_check.sh` and `validate_yaml.py` can also go away, but only after the app-interface ci.ext job is gone. So nbd if you want to clean up all three at the same time.)

Makefile
dustman9000: Thanks, I've updated this.

cmd/manager/main.go
```diff
@@ -34,10 +33,8 @@ import (
 
 // Change below variables to serve metrics on different host or port.
 var (
-	metricsHost                   = "0.0.0.0"
 	metricsPath                   = "/metrics"
 	metricsPort                   = "8080"
-	secretWatcherScanInterval     = time.Duration(10) * time.Minute
```
jharrington22, line 40: @dustman9000 are these changes intended?

pkg/clients/aws/route53.go
```diff
@@ -52,63 +52,6 @@ type awsClient struct {
 	client route53iface.Route53API
 }
 
-// searchForHostedZone finds a hostedzone when given an aws client and a domain string
```
jharrington22, line 55: @dustman9000 these too? There are a bunch of other edits after this that I wont comment on

jharrington22 (changes requested):

cmd/manager/main.go
jharrington22: Ah just saw the commit message, I figured they were from linting.

jharrington22 (approved):

Makefile
2uasimojo: :+1:

2uasimojo:
This is good to go once freeze is lifted.

/lgtm

2uasimojo:
/retest

Merge blocker closed.

yithian (approved):

PR #175, opened 2021-02-01
OSD-6070 patch clusterdeployment instead of update

mrbarge:
### What type of PR is this?
Refactor

### What this PR does / why we need it?
This PR changes the logic of the `certman-operator` to perform a `Patch()` of the `clusterdeployment` CR when adding or removing finalizers, rather than an `Update()`, in order to eliminate the possibility that the `Update()` call will drop data from the CR.

### Which Jira/Github issue(s) this PR fixes?
[OSD-6070](https://issues.redhat.com/browse/OSD-6070)

### Special notes for your reviewer:
I've tested this with a local deployment of `certman-operator` and verified that finalizers are still created and removed as expected. I have also verified that - if the Hive `clusterdeployment` CRD introduces a new field and sets that value in a `clusterdeployment` CR - the currently running operator does not remove that new field when it performs the `Patch()`, as per the goals of [OSD-6070](https://issues.redhat.com/browse/OSD-6070).

Related PRs:
- [pagerduty-operator](https://github.com/openshift/pagerduty-operator/pull/137)
- [deadmanssnitch-operator](https://github.com/openshift/deadmanssnitch-operator/pull/99)

### Pre-checks (if applicable):
- [X] Tested latest changes against a cluster

### DIFF
--- a/pkg/controller/certificaterequest/certificaterequest_controller.go
+++ b/pkg/controller/certificaterequest/certificaterequest_controller.go
@@ -145,8 +145,9 @@ func (r *ReconcileCertificateRequest) Reconcile(request reconcile.Request) (reco
 	if !utils.ContainsString(cr.ObjectMeta.Finalizers, certmanv1alpha1.CertmanOperatorFinalizerLabel) {
 		reqLogger.Info("adding finalizer to the certificate request")
 		localmetrics.IncrementCertRequestsCounter()
+		baseToPatch := client.MergeFrom(cr.DeepCopy())
 		cr.ObjectMeta.Finalizers = append(cr.ObjectMeta.Finalizers, certmanv1alpha1.CertmanOperatorFinalizerLabel)
-		if err := r.client.Update(context.TODO(), cr); err != nil {
+		if err := r.client.Patch(context.TODO(), cr, baseToPatch); err != nil {
 			reqLogger.Error(err, err.Error())
 			return reconcile.Result{}, err
 		}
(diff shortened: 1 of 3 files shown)


NautiluX (commented):
shouldn't this be done the same way when [removing the finalizer|https://github.com/openshift/certman-operator/blob/1faa0c3e45f8a12d7c68c57b1d2e1c58694c186a/pkg/controller/certificaterequest/certificaterequest_controller.go#L236]?

mrbarge:
> shouldn't this be done the same way when [removing the finalizer](https://github.com/openshift/certman-operator/blob/1faa0c3e45f8a12d7c68c57b1d2e1c58694c186a/pkg/controller/certificaterequest/certificaterequest_controller.go#L236)?

Thanks a lot for the review, @NautiluX . The original commit was focused solely on updates to the `ClusterDeployment`, but I agree that there should be consistency in how finalizers for `CertificateRequest`s are handled. I've addressed that in the latest commit.

cblecker:
/uncc @c-e-brumm @mbarnes 
/assign @dustman9000 

Would you mind having a look at this, Dustin?

dustman9000:
Hi @mrbarge - This looks good in general. Would it be possible to add a couple tests to `certificaterequest_controller_test.go` to test update vs patch?

mrbarge:
Thanks for the review @dustman9000 !

I've updated the clusterdeployment controller test suite to now verify either the presence (or lack) of the Certman finalizer in the resulting `clusterdeployment` for each of the tests that get run. I feel that this exercises the logic around the finalizer's addition or removal pretty well. Hopefully this is the sort of thing you were after!

dustman9000:
Awesome, thanks Matt!

PR #176, opened 2021-02-25
OSD-6498 recover certificaterequest status from secret

yithian:


yithian:
basically this is just introducing a new unit test to confirm that a certificaterequest can use a cert/key secret imported from a clustermigration

there was a lot of refactoring of tests to set up this test and make maintenance easier going forward

pkg/controller/certificaterequest/test_helpers.go
```diff
@@ -58,22 +60,46 @@ var certRequest = &certmanv1alpha1.CertificateRequest{
...
-var certSecret = &v1.Secret{
+var validCertSecret = &corev1.Secret{
 	ObjectMeta: metav1.ObjectMeta{
 		Namespace: testHiveNamespace,
 		Name:      testHiveSecretName,
 	},
+	Data: map[string][]byte{
+		// this is an absolutely garbage self-signed cert. it should not be used for
+		// anything ever
+		corev1.TLSCertKey: []byte(`-----BEGIN CERTIFICATE-----
```
dustman9000, line 83: I'm wondering if it would make better sense to store this as a test asset instead (for easier maintenance and to make the issued dates easier to see).

dustman9000:
Looks good in general and nothing major to block this. Just some TODOs to make some of the test assets more maintainable going forward.
/lgtm
/hold until Alex gets back

PR #179, opened 2021-03-26
Update origin-operator-registry image tag

cblecker:
fixes [OSD-6827](https://issues.redhat.com/browse/OSD-6827)

### DIFF
--- a/hack/app_sre_create_image_catalog.sh
+++ b/hack/app_sre_create_image_catalog.sh
@@ -95,7 +95,7 @@ REGISTRY_IMG="quay.io/app-sre/${_OPERATOR_NAME}-registry"
 DOCKERFILE_REGISTRY="Dockerfile.olm-registry"
 
 cat <<EOF > $DOCKERFILE_REGISTRY
-FROM quay.io/openshift/origin-operator-registry:4.5
+FROM quay.io/openshift/origin-operator-registry:latest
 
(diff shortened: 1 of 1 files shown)


yithian:
/lgtm

seems sane

hack/app_sre_create_image_catalog.sh
```diff
@@ -95,7 +95,7 @@ REGISTRY_IMG="quay.io/app-sre/${_OPERATOR_NAME}-registry"
 DOCKERFILE_REGISTRY="Dockerfile.olm-registry"
 
 cat <<EOF > $DOCKERFILE_REGISTRY
-FROM quay.io/openshift/origin-operator-registry:4.5
+FROM quay.io/openshift/origin-operator-registry:latest
```
NautiluX, line 98: I'd prefer using a specific tag instead of latest here for the following reasons:
* Using latest to avoid CVEs gives a false sense of security, as still a PR to the repo needs to be merged to have the CD create and deploy a new operator image.
* Unrelated PRs will use newer operator-registry images which may break the operator and debugging becomes a nightmare because it's not obvious the image changed.
Rather than setting this to latest on all operators, I'd prefer a pipeline that opens PRs when new images are released against all repositories. Until that's in place, l suggest we just use the latest currently available image (4.9 according to [this PR](https://github.com/openshift/gcp-project-operator/pull/134)).

PR #180, opened 2021-03-30
OSD-6777 cluster relocation during reconcile

yithian:


pkg/controller/certificaterequest/certificaterequest_controller.go
```diff
@@ -191,6 +186,15 @@ func (r *ReconcileCertificateRequest) Reconcile(request reconcile.Request) (reco
 		return reconcile.Result{}, err
 	}
 
+	// fetch the clusterdeployment and bail out if there's an outgoing migration annotation again
```
2uasimojo, line 189: I gather the idea is to check/bail before any write operation? If so, I don't think this block is necessary, since `ShouldReissue` is read-only, and you're doing another check right before `IssueCertificate`.

pkg/controller/certificaterequest/certificaterequest_controller.go
```diff
@@ -330,3 +343,20 @@ func (r *ReconcileCertificateRequest) revokeCertificateAndDeleteSecret(reqLogger
...
 	return nil
 }
+
+// relocationBailOut checks to see if there's a cluster relocation in progress
+func relocationBailOut(k client.Client, nsn types.NamespacedName) (relocating bool, err error) {
+	relocating = false
+
+	cd := &hivev1.ClusterDeployment{}
+	err = k.Get(context.TODO(), nsn, cd)
+
```
2uasimojo, line 353: I get that the loop below correctly iterates zero times if the `Get` failed, but it would be more defensive and explicit to include

```go
if err != nil {
    return
}
```

2uasimojo (commented):
Drive-by observation: If the main Reconcile were made more idempotent, this (multiple relocation checks) probably wouldn't be necessary. For example, if you requeued after adding the finalizer.

This PR seems like it's chasing timing issues, which is scary. What's the motivation behind it?

pkg/controller/certificaterequest/certificaterequest_controller.go
yithian: fair enough, clarity is always a good idea

pkg/controller/certificaterequest/certificaterequest_controller.go
yithian: nice catch! i had it in my head that `IssueCertificate` did more than it does

2uasimojo:
> The thought was to handle a situation where the clusterdeployment gets labelled to relocate to another hive instance in the middle of a reconcile loop. Granted that's pretty unlikely, but if and when we get up to the thousands of clusters, it doesn't seem like something impossible.

So, right, this was my point about chasing timing issues. You're narrowing the window by doing multiple checks, but you're not eliminating it. What if the label comes in between the check and `IssueCertificate`?

I haven't been involved in the design discussions around this concept, but:
- If there's truly a critical point where the next action is catastrophic if relocation has been requested, we need a more reliable synchronization mechanism.
- If there's not... then we really don't need to worry about it, and the one check you already had should be fine.

2uasimojo:
Also: Thanks for those updates, they look good to me. But I'll refrain from actually stamping this PR, since I'm not deep enough in the design.

pkg/controller/certificaterequest/certificaterequest_controller.go
```diff
@@ -146,17 +146,12 @@ func (r *ReconcileCertificateRequest) Reconcile(request reconcile.Request) (reco
...
 		return reconcile.Result{}, err
 	}
-
-	// bail out of the loop if there's an outgoing relocation annotation
-	for a, v := range cd.Annotations {
-		if a == hiveRelocationAnnotation && strings.Split(v, "/")[1] == hiveRelocationOutgoingValue {
-			return reconcile.Result{}, nil
-		}
+	if relocating {
+		return reconcile.Result{}, nil
```
dustman9000, line 154: Can we add a log message or update the status on the CR?

pkg/controller/certificaterequest/certificaterequest_controller.go
```diff
@@ -200,6 +195,15 @@ func (r *ReconcileCertificateRequest) Reconcile(request reconcile.Request) (reco
 		return reconcile.Result{}, err
 	}
 
+	// fetch the clusterdeployment and bail out if there's an outgoing migration annotation again
+	relocating, err = relocationBailOut(r.client, types.NamespacedName{Namespace: request.Namespace, Name: clusterDeploymentName})
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	if relocating {
+		return reconcile.Result{}, nil
```
dustman9000, line 204: Same suggestion here to update the CR status for better observability.

pkg/controller/certificaterequest/certificaterequest_controller.go
yithian: done

pkg/controller/certificaterequest/certificaterequest_controller.go
yithian: also done

PR #181, opened 2021-03-31
Cluster migration no cleanup

yithian:
if a cluster is being migrated between hive instances, we don't want to reconcile the clusterdeployment


yithian:
this is for https://issues.redhat.com/browse/OSD-6861

dustman9000:
thanks!
/lgtm

PR #186, opened 2021-04-08
OSD-6943 use ownerReference to get clusterdeployment, not string manipulation

yithian:


pkg/controller/certificaterequest/certificaterequest_controller.go
```diff
@@ -141,10 +141,12 @@ func (r *ReconcileCertificateRequest) Reconcile(request reconcile.Request) (reco
...
-	// deduce the clusterdeployment name from the certificate secret name
-	certificateSecretName := strings.Split(cr.Spec.CertificateSecret.Name, "-secret")[0] // find the bundle name from the certificate secret
-	requestNameSplit := fmt.Sprintf("-%s", certificateSecretName)                        // remove the certificate secret name from it
-	clusterDeploymentName := strings.Split(request.Name, requestNameSplit)[0]            // remove it from the certificate request name
+	clusterDeploymentName := ""
+	for _, ownerRef := range cr.OwnerReferences {
+		if ownerRef.Kind == "ClusterDeployment" {
+			clusterDeploymentName = ownerRef.Name
+		}
+	}
```
dustman9000, line 150: What if we are unable to get the clusterDeploymentName at this point? Throw an error?

dustman9000: Maybe a unit test around this?

yithian: i can make a unit test for that - they failed at first because of this anyway :laughing: 

what would it mean if there were no clusterdeployment owner reference for a certificaterequest? i'm not sure how things could get into that state without removing the finalizer on the clusterdeployment and then deleting it?

dustman9000:
thanks!
/lgtm

PR #199, opened 2021-06-30
Adding fake cluster support - OSD-7515

sulmer-redhat:
Skip certificate generation for fake clusters (ClusterDeployments labeled with the hive "fake" lable) as per [OSD-7515](https://issues.redhat.com/browse/OSD-7515)

### DIFF
--- a/pkg/controller/clusterdeployment/clusterdeployment_controller.go
+++ b/pkg/controller/clusterdeployment/clusterdeployment_controller.go
@@ -48,10 +48,11 @@ import (
 var log = logf.Log.WithName("controller_clusterdeployment")
 
 const (
-	controllerName                = "clusterdeployment"
-	ClusterDeploymentManagedLabel = "api.openshift.com/managed"
-	hiveRelocationAnnotation      = "hive.openshift.io/relocate"
-	hiveRelocationOutgoingValue   = "outgoing"
+	controllerName                  = "clusterdeployment"
+	ClusterDeploymentManagedLabel   = "api.openshift.com/managed"
+	hiveRelocationAnnotation        = "hive.openshift.io/relocate"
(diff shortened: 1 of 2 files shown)


pkg/controller/clusterdeployment/clusterdeployment_controller.go
```diff
@@ -48,10 +48,11 @@ import (
...
 const (
-	controllerName                = "clusterdeployment"
-	ClusterDeploymentManagedLabel = "api.openshift.com/managed"
-	hiveRelocationAnnotation      = "hive.openshift.io/relocate"
-	hiveRelocationOutgoingValue   = "outgoing"
+	controllerName                  = "clusterdeployment"
+	ClusterDeploymentManagedLabel   = "api.openshift.com/managed"
+	hiveRelocationAnnotation        = "hive.openshift.io/relocate"
+	hiveRelocationOutgoingValue     = "outgoing"
+	fakeClusterDeploymentAnnotation = "hive.openshift.io/fake-cluster"
```
ramonbutter, line 55: can we maybe import that constant from the hive package?
Just to not repeat ourself
https://github.com/openshift/hive/blob/697460baf215c1231db81329d57311d137fad456/pkg/constants/constants.go#L321

ramonbutter: But I see we are doing also for other hive labels...

yithian: that's a good idea. i wasn't aware of that package when i put `hiveRelocationAnnotation` and `hiveRelocationOutgoingValue` in here originally

yithian:
/lgtm

@sulmer-redhat feel free to throw another change in, pulling the hive labels from the hive constants package, otherwise we can deal with it later

PR #201, opened 2021-07-21
OSD-7084 alert when cert will expire

yithian:
this enables metric scraping by prometheus and includes updates to the local testing tooling


deploy/service.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
```
dustman9000, line 1: I think we should probably move the Service and ServiceMonitor out into the hack/test/deploy directory instead as this is handled automatically be the operator-sdk/olm

hack/test/clusterdeploy.yaml
```diff
@@ -14,69 +14,40 @@ metadata:
   namespace: certtest
 spec:
   installed: true
-  baseDomain: dnstest.aws.openshift.com
+  baseDomain: dev.yuggoth.space
```
dustman9000, line 17: Maybe make generic like app.example.com?

dustman9000: Can we move this file into something like hack/test/deploy?

hack/test/local_test.sh
```diff
@@ -2,7 +2,7 @@
 # This script will test certman-operator in minikube by doing the following:
 # 1. Create a test cluster locally.
 # 2. Install certman's dependencies and CRDs.
-# 3. Create a docker image of the operator.
+# 3. Create a podman image of the operator.
```
dustman9000, line 5:
```suggestion
# 3. Create a podman/docker image of the operator.
```

hack/test/local_test.sh
```diff
@@ -66,23 +66,24 @@ if [[ ! $(pwd) =~ .*certman-operator$ ]]; then
   exit
 fi
 
-echo "Checking if docker service is active"
-systemctl is-active docker
+#echo "Checking if docker service is active"
```
dustman9000, line 69: Can we support both podman and docker in this script?

hack/test/local_test.sh
```diff
@@ -66,23 +66,24 @@ if [[ ! $(pwd) =~ .*certman-operator$ ]]; then
...
+#systemctl is-active docker
 
 testdir="${initial_dir}/hack/test"
 tmpdir="${initial_dir}/hack/test/tmp"
 mkdir ${tmpdir}
 
 # Start minikube with extra memory (needed for the go build)
-minikube start -p certtest --memory='2500mb'
+minikube start -p certtest --memory=6g --bootstrapper=kubeadm --extra-config=kubelet.authentication-token-webhook=true --extra-config=kubelet.authorization-mode=Webhook --extra-config=scheduler.address=0.0.0.0 --extra-config=controller-manager.address=0.0.0.0 --extra-config=apiserver.service-node-port-range=1-65535
```
dustman9000, line 77: Move extra params into a variable to make it more maintainable?

hack/test/local_test.sh
```diff
@@ -66,23 +66,24 @@ if [[ ! $(pwd) =~ .*certman-operator$ ]]; then
...
 kubectl config use-context certtest
 
 # Install openshift router
 cd $tmpdir
 git clone git@github.com:openshift/router.git
 cd router
-kubectl create ns openshift-ingress
-kubectl create -n openshift-ingress -f deploy/
+kubectl create -n openshift-ingress -f deploy/route_crd.yaml
```
dustman9000, line 84: Can we move this into a hack/test/deploy directory?

hack/test/namespace.yaml
```diff
@@ -3,6 +3,8 @@ apiVersion: v1
 kind: Namespace
```
dustman9000, line 3: ditto into a hack/test/deploy directory

hack/test/clusterdeploy.yaml
yithian: ugh, i thought i excluded that change. i'll revert hostname change and move the clusterdeployment into hack/test/deploy.yml

hack/test/local_test.sh
yithian: that specific resource is from the openshift/router git repo, so it's not something we have to worry about

hack/test/local_test.sh
yithian: that should be added now

PR #204, opened 2021-09-03
OSD-8263 recover a missing OwnerReference for a CertificateRequest

yithian:
this assumes that:
* there will be exactly 1 ClusterDeployment per namespace that the
  CertificateRequest is in
* the CertificateRequest request is named
  <ClusterDeployment.Name>--primary-cert-bundle

### DIFF
--- a/pkg/controller/certificaterequest/certificaterequest_controller.go
+++ b/pkg/controller/certificaterequest/certificaterequest_controller.go
@@ -18,6 +18,7 @@ package certificaterequest
 
 import (
 	"context"
+	gerrors "errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


pkg/controller/certificaterequest/certificaterequest_controller.go
```diff
@@ -143,20 +142,38 @@ func (r *ReconcileCertificateRequest) Reconcile(request reconcile.Request) (reco
...
 	}
 
-	clusterDeploymentName := ""
-	for _, ownerRef := range cr.OwnerReferences {
-		if ownerRef.Kind == "ClusterDeployment" {
-			clusterDeploymentName = ownerRef.Name
-		}
+	// assume there's only one clusterdeployment in a namespace and that it's the owner of this certificaterequest
+	// we have to assume this so that if/when a CertificateRequest loses its OwnerReferences, it can still reconcile
+	clusterDeploymentName := strings.Split(request.NamespacedName.Name, certRequestSuffix)[0]
```
dustman9000, line 147: Can we add a defensive check? What if it's nil? We'll have a nil deref.

PR #205, opened 2021-10-03
Fixed a bug with `certman_operator_certificate_valid_duration_days`

qhua948:
Fixed the bug where UpdateCertValidDuration returns the wrong results if
the two dates are in different years


yithian:
/lgtm 

thanks!

PR #208, opened 2021-12-03
prepare a FEDRAMP env variable for use

yithian:
for [OSD-0249](https://issues.redhat.com/browse/OSD-0249)

### DIFF
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -11,7 +11,7 @@ parameters:
 - name: CHANNEL
   value: staging
 - name: FEDRAMP
-  required: true
+  value: "false"
 
 objects:
 - apiVersion: operators.coreos.com/v1alpha1
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -11,7 +11,7 @@ parameters:
 - name: CHANNEL
   value: staging
 - name: FEDRAMP
-  value: "false"
+  value: "${FEDRAMP}"
 
 objects:
 - apiVersion: operators.coreos.com/v1alpha1
--- a/hack/olm-registry/olm-artifacts-template.yaml
+++ b/hack/olm-registry/olm-artifacts-template.yaml
@@ -10,6 +10,8 @@ parameters:
   required: true
 - name: CHANNEL
   value: staging
+- name: FEDRAMP
+  required: true
 
 objects:
 - apiVersion: operators.coreos.com/v1alpha1
--- a/hack/test/deploy/deploy.yaml
+++ b/hack/test/deploy/deploy.yaml
@@ -33,4 +33,4 @@ spec:
             - name: OPERATOR_NAME
               value: "certman-operator"
             - name: FEDRAMP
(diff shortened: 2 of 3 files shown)


yithian:
the lint failure is because the fedramp variable isn't being used yet

hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -10,6 +10,8 @@ parameters:
   required: true
 - name: CHANNEL
   value: staging
+- name: FEDRAMP
+  required: true
```
maorfr, line 14: i recommend not making this required, and set the default value to be `"false"`.

hack/test/deploy/deploy.yaml
```diff
@@ -32,3 +32,5 @@ spec:
                   fieldPath: metadata.name
             - name: OPERATOR_NAME
               value: "certman-operator"
+            - name: FEDRAMP
+              value: false
```
maorfr, line 36: value should probably be `"false"`

hack/test/deploy/deploy.yaml
```diff
@@ -32,3 +32,5 @@ spec:
                   fieldPath: metadata.name
             - name: OPERATOR_NAME
               value: "certman-operator"
+            - name: FEDRAMP
+              value: "false"
```
maorfr, line 36: `value: "${FEDRAMP}"`?

pkg/controller/certificaterequest/certificaterequest_controller.go
```diff
@@ -55,6 +56,7 @@ const (
 	certRequestSuffix                     = "-primary-cert-bundle"
 )
 
+var fedramp = os.Getenv("FEDRAMP") == "true"
```
dustman9000, line 59: Can we add a `log.Info` message for the presence of this var? Also, should "FEDRAMP" be a const?

yithian: i tried defining it as a const but that threw an error: `os.Getenv("FEDRAMP") == "true" (untyped bool value) is not constant`

pkg/controller/certificaterequest/certificaterequest_controller.go
```diff
@@ -118,6 +120,11 @@ func (r *ReconcileCertificateRequest) Reconcile(request reconcile.Request) (reco
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 
 	reqLogger.Info("reconciling CertificateRequest")
+	if len(os.Getenv("FEDRAMP")) > 0 {
```
dustman9000, line 123: Could this just be the `fedramp` var?

hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -10,6 +10,8 @@ parameters:
   required: true
 - name: CHANNEL
   value: staging
+- name: FEDRAMP
+  value: "${FEDRAMP}"
```
maorfr, line 14: This needs to be "false"

yithian: https://github.com/openshift/certman-operator/pull/209 opened to address this

PR #209, opened 2021-12-06
Revert "pull FEDRAMP value from environment"

yithian:
This reverts commit 50e54cd315d3c667797f638e2369d20a1f40fced to address https://github.com/openshift/certman-operator/pull/208#discussion_r763276137


maorfr:
Looks good. Where is the value being used though?

maorfr:
Maybe this will help? https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/subscription-config.md

PR #210, opened 2021-12-07
OSD-9241 add support for ZONE_ID deployment param

yithian:


hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -11,7 +11,13 @@ parameters:
 - name: CHANNEL
   value: staging
 - name: FEDRAMP
+<<<<<<< HEAD
```
dustman9000, line 14: Looks like a merge conflict

yithian: whoops, good catch, thanks

deploy/operator.yaml
```diff
@@ -32,3 +32,5 @@ spec:
               value: "rh-api"
             - name: FEDRAMP
               value: "false"
+            - name: ZONE_ID
```
dustman9000, line 35: Let's rename this to HOSTED_ZONE_ID to avoid confusion...

hack/olm-registry/olm-artifacts-template.yaml
```diff
@@ -45,3 +47,5 @@ objects:
       env:
       - name: FEDRAMP
         value: "${FEDRAMP}"
+      - name: HOSTED_ZONE_ID
+        value: ${ZONE_ID}
```
dustman9000, line 51: Should be HOSTED_ZONE_ID
```suggestion
        value: ${HOSTED_ZONE_ID}
```

pkg/controller/certificaterequest/certificaterequest_controller.go
```diff
@@ -126,6 +126,12 @@ func (r *ReconcileCertificateRequest) Reconcile(request reconcile.Request) (reco
 		reqLogger.Info("running in FedRAMP environment: %b", fedramp)
 	}
 
+	var fedramp_hosted_zone_id string
+	if fedramp {
+		fedramp_hosted_zone_id = os.Getenv("HOSTED_ZONE_ID")
+		reqLogger.Info("running in FedRAMP zone: %s", fedramp_hosted_zone_id)
```
dustman9000, line 132: Check for zero length string?

dustman9000 (changes requested):
Need to fix the parameter expansion from ZONE_ID to HOSTED_ZONE_ID

PR #212, opened 2022-01-11
[OSD-8748] Update route53 client to support fedramp DNS01 challenges

blrm:
This updates the route53 client to follow the DNS01 challenge flow defined for fedramp clusters as described in https://issues.redhat.com/browse/OSD-8748. It checks for the presence of a FEDRAMP env var, and when present and set to True, it creates the DNS01 challenge records under a single public hosted zone in a non-govcloud account. This allows us to work around the fact that govcloud route53 only allows for private hosted zones.

### DIFF
--- a/pkg/clients/aws/route53.go
+++ b/pkg/clients/aws/route53.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"strings"
 	"time"
 
@@ -47,6 +48,8 @@ const (
 	awsCredsSecretIDKey         = "aws_access_key_id"
 	awsCredsSecretAccessKey     = "aws_secret_access_key"
 	awsCredsSecretName          = "certman-operator-aws-credentials"
+	fedrampEnvVariable          = "FEDRAMP"
+	fedrampHostedZoneIDVariable = "HOSTED_ZONE_ID"
 	resourceRecordTTL           = 60
 	clientMaxRetries            = 25
 	retryerMaxRetries           = 10
@@ -56,6 +59,9 @@ const (
 	clusterDeploymentSTSLabel   = "api.openshift.com/sts"
 	configMapSTSJumpRoleField   = "sts-jump-role"
 )
+
+var fedramp = os.Getenv(fedrampEnvVariable) == "true"
+var fedrampHostedZoneID = os.Getenv(fedrampHostedZoneIDVariable)
 
 // awsClient implements the Client interface
 type awsClient struct {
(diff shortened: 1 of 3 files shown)


pkg/clients/aws/route53.go
```diff
@@ -57,6 +60,9 @@ const (
 	configMapSTSJumpRoleField   = "sts-jump-role"
 )
 
+var fedramp = os.Getenv(fedrampEnvVariable) == "true"
```
dustman9000, line 63: I think os.LookupEnv() would be a little safer here as it handles empty/nil values better.
https://pkg.go.dev/os#LookupEnv

pkg/clients/aws/route53.go
```diff
@@ -70,6 +76,45 @@ func (c *awsClient) AnswerDNSChallenge(reqLogger logr.Logger, acmeChallengeToken
 	fqdn = fmt.Sprintf("%s.%s", cTypes.AcmeChallengeSubDomain, domain)
 	reqLogger.Info(fmt.Sprintf("fqdn acme challenge domain is %v", fqdn))
 
+	if fedramp {
+		zone, err := c.client.GetHostedZone(&route53.GetHostedZoneInput{Id: &fedrampHostedZoneID})
+		if err != nil {
+			return "", err
```
dustman9000, line 82: Can we add a reqLogger.Error() here?

pkg/clients/aws/route53.go
```diff
@@ -132,8 +177,63 @@ func (c *awsClient) AnswerDNSChallenge(reqLogger logr.Logger, acmeChallengeToken
...
 // and attempts to write a test TXT ResourceRecord to it. If successful, will return `true, nil`.
 func (c *awsClient) ValidateDNSWriteAccess(reqLogger logr.Logger, cr *certmanv1alpha1.CertificateRequest) (bool, error) {
-	hostedZones, err := listAllHostedZones(c.client, &route53.ListHostedZonesInput{})
+
+	var hostedZones []*route53.HostedZone
+	var err error
+	if fedramp {
+		zone, err := c.client.GetHostedZone(&route53.GetHostedZoneInput{Id: &fedrampHostedZoneID})
+		if err != nil {
+			return false, err
```
dustman9000, line 186: Can we add a reqLogger.Error() here?

pkg/controller/certificaterequest/certificaterequest_controller.go
```diff
@@ -126,15 +126,15 @@ func (r *ReconcileCertificateRequest) Reconcile(request reconcile.Request) (reco
 	if len(os.Getenv(fedrampEnvVariable)) == 0 {
```
dustman9000, line 126: Since you're already in this code, can we change this to use os.LookupEnv instead?

PR #216, opened 2022-02-17
Added output of response status from cloudflare

aliceh:
In this commit, I added outputting the response status from Cloudflare with the status code into the logs; a custom local metric to count DNS failures. I added a call to google's public DNS server in case the call to Cloudflare's public DNS server fails.

### DIFF
--- a/pkg/controller/certificaterequest/cloudflare.go
+++ b/pkg/controller/certificaterequest/cloudflare.go
@@ -25,31 +25,32 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	"github.com/openshift/certman-operator/pkg/localmetrics"
 	"github.com/pkg/errors"
 )
 
-type CloudflareQuestion struct {
+type DnsServerQuestion struct {
 	Name string `json:"name"`
 	Type int    `json:"type"`
 }
 
-type CloudflareAnswer struct {
+type DnsServerAnswer struct {
 	Name string `json:"name"`
 	Type int    `json:"type"`
 	TTL  int    `json:"TTL"`
 	Data string `json:"data"`
 }
 
-type CloudflareResponse struct {
-	Status    int                  `json:"Status"`
-	TC        bool                 `json:"TC"`
-	RC        bool                 `json:"RC"`
-	RA        bool                 `json:"RA"`
-	AD        bool                 `json:"AD"`
-	CD        bool                 `json:"CD"`
-	Questions []CloudflareQuestion `json:"Question"`
-	Answers   []CloudflareAnswer   `json:"Answer"`
-	Authority []CloudflareAnswer   `json:"Authority"`
+type DnsServerResponse struct {
+	Status    int                 `json:"Status"`
+	TC        bool                `json:"TC"`
+	RC        bool                `json:"RC"`
+	RA        bool                `json:"RA"`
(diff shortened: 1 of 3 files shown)


pkg/controller/certificaterequest/cloudflare.go
```diff
@@ -152,6 +152,20 @@ func FetchResourceRecordUsingCloudflareDNS(reqLogger logr.Logger, name string) (
 		return nil, err
 	}
 
+	if 400 <= response.StatusCode && response.StatusCode <= 499 {
+
```
yithian, line 156:
```suggestion
```

pkg/controller/certificaterequest/cloudflare.go
```diff
@@ -152,6 +152,20 @@ func FetchResourceRecordUsingCloudflareDNS(reqLogger logr.Logger, name string) (
...
+		reqLogger.Info("Client Error: " + response.Status)
+
```
yithian, line 158:
```suggestion
```

pkg/controller/certificaterequest/cloudflare.go
```diff
@@ -152,6 +152,20 @@ func FetchResourceRecordUsingCloudflareDNS(reqLogger logr.Logger, name string) (
...
+	} else if 500 <= response.StatusCode && response.StatusCode <= 599 {
+
```
yithian, line 160:
```suggestion
```

pkg/controller/certificaterequest/cloudflare.go
```diff
@@ -152,6 +152,20 @@ func FetchResourceRecordUsingCloudflareDNS(reqLogger logr.Logger, name string) (
...
+		reqLogger.Info("Server Error: " + response.Status)
+
```
yithian, line 162:
```suggestion
```

pkg/controller/certificaterequest/cloudflare.go
```diff
@@ -152,6 +152,20 @@ func FetchResourceRecordUsingCloudflareDNS(reqLogger logr.Logger, name string) (
...
+	} else {
+
```
yithian, line 164:
```suggestion
```

pkg/controller/certificaterequest/cloudflare.go
```diff
@@ -152,6 +152,20 @@ func FetchResourceRecordUsingCloudflareDNS(reqLogger logr.Logger, name string) (
...
+		reqLogger.Info("response Status from Cloudflare: " + response.Status)
+
```
yithian, line 166:
```suggestion
```

yithian (changes requested):
let's clean up some of this whitespace in the if/else block, otherwise this seems sane

yithian (changes requested):
i'd recommend just using `cloudflareDNSOverHttpsEndpoint` or `googleDNSOverHttpsEndpoint` directly in the `FetchResourceRecordUsingPublicDNS()` calls instead of using an intermediary variable (`dnsOverHttpsEndpoint`)--i don't think it adds anything. did i miss something about how that variable is used?

otherwise, this looks good!

yithian:
/lgtm

thanks for working through all my suggestions :sweat_smile:

PR #218, opened 2022-03-09
OSD-1949 add a rate limiter to the certificaterequest controller

yithian:
this should stop certman-operator from burning through the rate limit of AWS and let's encrypt


sam-nguyen7 (approved):
/lgtm

PR #219, opened 2022-03-15
removes support for deprecated LE secret names

yithian:
the secret names lets-encrypt-account-staging and lets-encrypt-account-production have been deprecated for a while now. this commit removes support for using those names, only looking for the standard secret name: lets-encrypt-account

### DIFF
--- a/pkg/leclient/lets_encrypt.go
+++ b/pkg/leclient/lets_encrypt.go
@@ -27,7 +27,6 @@ import (
 
 	"github.com/eggsampler/acme"
 	v1 "k8s.io/api/core/v1"
-	kerr "k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/openshift/certman-operator/config"
@@ -233,18 +232,6 @@ func getLetsEncryptAccountSecret(kubeClient client.Client) (secret *v1.Secret, e
 	secretName := letsEncryptAccountSecretName
 
 	secret, err = GetSecret(kubeClient, secretName, config.OperatorNamespace)
-	if err != nil {
-		// If it's not found err, try to use the legacy production secret name for backward compatibility
-		if kerr.IsNotFound(err) {
-			secretName = letsEncryptProductionAccountSecretName
-			secret, err = GetSecret(kubeClient, secretName, config.OperatorNamespace)
-			// If it's not found err, try to use the legacy staging secret name for backward compatibility
-			if kerr.IsNotFound(err) {
-				secretName = letsEncryptStagingAccountSecretName
-				secret, err = GetSecret(kubeClient, secretName, config.OperatorNamespace)
-			}
-		}
-	}
 	return
 }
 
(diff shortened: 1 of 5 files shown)


pkg/leclient/lets_encrypt.go
```diff
@@ -233,18 +232,6 @@ func getLetsEncryptAccountSecret(kubeClient client.Client) (secret *v1.Secret, e
 	secretName := letsEncryptAccountSecretName
 
 	secret, err = GetSecret(kubeClient, secretName, config.OperatorNamespace)
-	if err != nil {
```
sam-nguyen7, line 236: so we're not checking if this errors out at all? is this intentional?

sam-nguyen7 (commented):
Looks great! @yithian would you mind writing a quick description of this PR as well? thank you!

pkg/leclient/lets_encrypt.go
yithian: correct - i don't think there's any error handling to be done here, so any error will be returned to where `getLetsEncryptAccountSecret()` is called for handling at that point

if there's some handling we can do at this point, i'm happy to add it though

yithian:
> Looks great! @yithian would you mind writing a quick description of this PR as well? thank you!

description updated

pkg/leclient/lets_encrypt.go
```diff
@@ -233,18 +232,6 @@ func getLetsEncryptAccountSecret(kubeClient client.Client) (secret *v1.Secret, e
...
-			secretName = letsEncryptProductionAccountSecretName
-			secret, err = GetSecret(kubeClient, secretName, config.OperatorNamespace)
-			// If it's not found err, try to use the legacy staging secret name for backward compatibility
-			if kerr.IsNotFound(err) {
-				secretName = letsEncryptStagingAccountSecretName
-				secret, err = GetSecret(kubeClient, secretName, config.OperatorNamespace)
-			}
-		}
-	}
 	return
```
sam-nguyen7, line 235: what is this returning?

sam-nguyen7 (changes requested):

pkg/leclient/lets_encrypt.go
yithian: since `getLetsEncryptAccountSecret()` sets up named return variables `(secret *v1.Secret, err error)`, a naked return will return those specific variables: https://go.dev/tour/basics/7

but you're right, this function doesn't do much anymore. i'll axe it

pkg/leclient/lets_encrypt.go
```diff
@@ -233,18 +232,6 @@ func getLetsEncryptAccountSecret(kubeClient client.Client) (secret *v1.Secret, e
 	secretName := letsEncryptAccountSecretName
 
 	secret, err = GetSecret(kubeClient, secretName, config.OperatorNamespace)
```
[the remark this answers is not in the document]
yithian: removed the entire function

sam-nguyen7 (approved):
/lgtm

PR #220, opened 2022-03-29
OSD-10349 better tests

yithian:
improve test coverage


yithian:
i'm opening this PR early for ease of tracking test coverage across the whole project rather than file-by-file, the way `go test ./...` does

fahlmant (changes requested):
Please provide a description

PR #231, opened 2023-03-30
erroring during DNS record removal on GCP

anispate:
Part of [OSD-13267](https://issues.redhat.com/browse/OSD-13267) ticket.

The "Route53" message is a red herring, it's not actually trying to do Route53 things. Therefore, made that message common across cloud-provider.

But the GCP error seems like upstream issue.


pkg/clients/gcp/dns.go
```diff
@@ -147,10 +148,16 @@ func (c *gcpClient) DeleteAcmeChallengeResourceRecords(reqLogger logr.Logger, cr
 
 	// clean up records
 	err = c.deleteDnsRecords(zone, changes)
+
 	if err != nil {
-		return err
+		if e, ok := err.(*googleapi.Error); ok {
+			if e.Code == 400 && e.Message == "The 'entity.change' parameter is required but was missing." {
```
cblecker, line 154: Ignoring this error means we aren't deleting these records.

PR #232, opened 2023-03-31
Bump go dependencies

cblecker:


go.mod
```diff
@@ -3,107 +3,103 @@ module github.com/openshift/certman-operator
...
-	github.com/openshift/api v0.0.0-20220531073726-6c4f186339a7
-	github.com/openshift/aws-account-operator/api v0.0.0-20220919121610-00e208e977e8
-	github.com/openshift/hive/apis v0.0.0-20220831193253-7dca200fad2f
-	github.com/openshift/operator-custom-metrics v0.5.1-0.20220802235640-dc76a1f15ee8
+	github.com/go-logr/logr v1.2.4
+	github.com/lib/pq v1.10.7
+	github.com/openshift/api v0.0.0-20230228142948-d170fcdc0fa6
+	github.com/openshift/aws-account-operator/api v0.0.0-20230322125717-5b5a00a3e99f
+	github.com/openshift/hive/apis v0.0.0-20230329164219-869f3fac05bd
+	github.com/openshift/operator-custom-metrics v0.5.0
```
mjlshen, line 17: I double-checked this dep and I think it's ok for certman to go backwards here. Originally it was pinned to a commit to pick up this fix https://github.com/openshift/operator-custom-metrics/pull/44 like the rest of our hive operators, so they wouldn't need to explicity register routes during setup, however I see that it registers routes anyway https://github.com/openshift/certman-operator/blob/master/main.go#L79.

go.mod
```diff
@@ -3,107 +3,103 @@ module github.com/openshift/certman-operator
 go 1.17
```
mjlshen, line 3: Since boilerplate supports Go 1.19 any reason not to bump this too? I'm ok with separating it into a separate PR too if we're just trying to break apart changes in to smaller pieces.

go.mod
cblecker: Hmm good catch. Should we release a new version of operator-customer-metrics? cc @fahlmant

go.mod
cblecker: Yeah let's do that in a follow up. Thanks for pointing it out

go.mod
cblecker: Bumped to 0.5.1 to pick up all the latest changes

go.mod
cblecker: On second thought, let's do it.

PR #261, opened 2024-02-13
Fixing Fix the certman operator logic to look up HZ ID by DNSZone.Status.AWS.ZoneID in Hive

anispate:
[OSD-17701](https://issues.redhat.com/browse/OSD-17701) - Fix the certman operator logic to look up HZ ID by DNSZone.Status.AWS.ZoneID in Hive.

### DIFF
--- a/controllers/certificaterequest/issue_certificate.go
+++ b/controllers/certificaterequest/issue_certificate.go
@@ -231,10 +231,10 @@ func (r *CertificateRequestReconciler) FindZoneIDForChallenge(namespace string) 
 	}
 
 	if *dnsZones.Items[0].Status.AWS.ZoneID != "" {
-		return filepath.Base(*dnsZones.Items[0].Status.AWS.ZoneID), nil
-	} else if *dnsZones.Items[0].Status.GCP.ZoneName != "" {
-		return *dnsZones.Items[0].Status.GCP.ZoneName, nil
-	} else {
-		return "", nil
+		return filepath.Base(*dnsZones.Items[0].Status.AWS.ZoneID), nil //the format of that ID is (/hostedzone/<something>) so, we don't care about >/. Here, we only want to grab the stuff after the last
 	}
+	if *dnsZones.Items[0].Status.GCP.ZoneName != "" {
+		return *dnsZones.Items[0].Status.GCP.ZoneName, nil
+	}
+	return "", err
 }
--- a/controllers/certificaterequest/issue_certificate.go
+++ b/controllers/certificaterequest/issue_certificate.go
@@ -17,6 +17,7 @@ limitations under the License.
 package certificaterequest
 
 import (
+	"context"
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/x509"
@@ -25,15 +26,17 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"path/filepath"
 	"strings"
 
 	"github.com/go-logr/logr"
-	"github.com/prometheus/client_golang/prometheus"
-	corev1 "k8s.io/api/core/v1"
-
 	certmanv1alpha1 "github.com/openshift/certman-operator/api/v1alpha1"
 	"github.com/openshift/certman-operator/pkg/leclient"
 	"github.com/openshift/certman-operator/pkg/localmetrics"
+	hivev1 "github.com/openshift/hive/apis/hive/v1"
+	"github.com/prometheus/client_golang/prometheus"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
@@ -110,8 +113,13 @@ func (r *CertificateRequestReconciler) IssueCertificate(reqLogger logr.Logger, c
 		if keyAuthErr != nil {
 			return fmt.Errorf("Could not get authorization key for dns challenge")
 		}
-		fqdn, err := dnsClient.AnswerDNSChallenge(reqLogger, DNS01KeyAuthorization, domain, cr)
 
+		dnsZone, err := r.FindZoneIDForChallenge(cr.Namespace)
+		if err != nil {
+			return err
+		}
+
+		fqdn, err := dnsClient.AnswerDNSChallenge(reqLogger, DNS01KeyAuthorization, domain, cr, dnsZone)
 		if err != nil {
 			return err
 		}
@@ -209,4 +217,24 @@ func (r *CertificateRequestReconciler) IssueCertificate(reqLogger logr.Logger, c
 	}
 
 	return nil
+}
+
+func (r *CertificateRequestReconciler) FindZoneIDForChallenge(namespace string) (dnsZoneID string, err error) {
+	dnsZones := hivev1.DNSZoneList{}
+	err = r.Client.List(context.TODO(), &dnsZones, &client.ListOptions{Namespace: namespace})
+	if err != nil {
+		return "", err
+	}
+
+	if len(dnsZones.Items) != 1 {
+		return "", fmt.Errorf("%d dnsZone objects in a specific namespace found, expected 1 dnsZone", len(dnsZones.Items))
+	}
+
+	if *dnsZones.Items[0].Status.AWS.ZoneID != "" {
+		return filepath.Base(*dnsZones.Items[0].Status.AWS.ZoneID), nil
+	} else if *dnsZones.Items[0].Status.GCP.ZoneName != "" {
+		return *dnsZones.Items[0].Status.GCP.ZoneName, nil
+	} else {
+		return "", nil
+	}
 }
--- a/pkg/clients/aws/route53.go
+++ b/pkg/clients/aws/route53.go
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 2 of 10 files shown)


iamkirkbater:
/approve

I'll give others a chance to look before merging.

pkg/clients/aws/route53.go
```diff
@@ -66,7 +67,9 @@ var fedrampHostedZoneID = os.Getenv(fedrampHostedZoneIDVariable)
 
 // awsClient implements the Client interface
 type awsClient struct {
-	client route53iface.Route53API
+	client     route53iface.Route53API
+	kubeClient client.Client
+	namespace  string
```
fahlmant, line 72: These are unrelated to AWS, so should not be placed inside the AWS Client

pkg/clients/aws/route53.go
```diff
@@ -66,7 +67,9 @@ var fedrampHostedZoneID = os.Getenv(fedrampHostedZoneIDVariable)
 
 // awsClient implements the Client interface
 type awsClient struct {
-	client route53iface.Route53API
+	client     route53iface.Route53API
+	kubeClient client.Client
```
fahlmant, line 71: KubeClient and Namespace are unrelated to AWS, so should not be placed inside the AWS Client

fahlmant (changes requested):
I think the best way forward here to keep things untangled is to refactor into multiple functions
- FindZoneIDForChallenge
This function can handle determining which Zone ID is the right one for a cluster. This is where a kubeclient and the right namespace can be passed as arguments to the function and can be used to find the right hive resource to find which zone to use
- AnswerDNSChallenge
This function keeps the logic of creating the right Resource Record Set and creating it in the target zone, but all the logic for determining which zone to use is removed. It takes the zone ID from FindZoneIDForChallenge as an argument.

This way, the packaging of the kube client and secret don't have to be bundled in with the AWS client, and the logic that interacts with AWS doesn't have to be crossed with the logic that interacts with the kubeapi.

controllers/certificaterequest/issue_certificate.go
```diff
@@ -210,3 +218,23 @@ func (r *CertificateRequestReconciler) IssueCertificate(reqLogger logr.Logger, c
...
+		return "", fmt.Errorf("%d dnsZone objects in a specific namespace found, expected 1 dnsZone", len(dnsZones.Items))
+	}
+
+	if *dnsZones.Items[0].Status.AWS.ZoneID != "" {
+		return filepath.Base(*dnsZones.Items[0].Status.AWS.ZoneID), nil
+	} else if *dnsZones.Items[0].Status.GCP.ZoneName != "" {
+		return *dnsZones.Items[0].Status.GCP.ZoneName, nil
+	} else {
+		return "", nil
+	}
```
fahlmant, line 239:
```suggestion
	} 
	
	return "", nil

```

fahlmant: Should this return an error if no Zones are found?

fahlmant (commented):
The fedramp and non-fedramp calls to `ChangeResourceRecordSets` look very similar. What's the difference between them, and can we combine them into one block?

controllers/certificaterequest/issue_certificate.go
anispate: Yes, you could be right that we can return an error but, I wasn't sure how to perform check ZoneID for the Azure so, the else statement was for Azure.

fahlmant: Does certman-operator support Azure? If so, then you'll need to implement the proper check. If not, then you can return the error and ignore Azure checks.

controllers/certificaterequest/issue_certificate.go
```diff
@@ -210,3 +218,23 @@ func (r *CertificateRequestReconciler) IssueCertificate(reqLogger logr.Logger, c
...
+	err = r.Client.List(context.TODO(), &dnsZones, &client.ListOptions{Namespace: namespace})
+	if err != nil {
+		return "", err
+	}
+
+	if len(dnsZones.Items) != 1 {
+		return "", fmt.Errorf("%d dnsZone objects in a specific namespace found, expected 1 dnsZone", len(dnsZones.Items))
+	}
+
+	if *dnsZones.Items[0].Status.AWS.ZoneID != "" {
```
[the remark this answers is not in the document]
fahlmant: there are more than two possible outcomes
1. AWS is there
2. GCP is there
3. Maybe Azure is there? (according to above comment)
4. None of the above (return err)

controllers/certificaterequest/issue_certificate.go
anispate: https://github.com/openshift/hive/blob/master/apis/hive/v1/dnszone_types.go#L149-L150 <<<< Seems like AzureDNSZoneStatus struct is empty so, We don't need Azure condition.

controllers/certificaterequest/issue_certificate.go
```diff
@@ -210,3 +218,22 @@ func (r *CertificateRequestReconciler) IssueCertificate(reqLogger logr.Logger, c
...
+		return filepath.Base(*dnsZones.Items[0].Status.AWS.ZoneID), nil
```
fahlmant, line 234: Why is `filepath.Base` needed here?

controllers/certificaterequest/issue_certificate.go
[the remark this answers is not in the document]
rafael-azevedo: @anispate I am not sure you addressed the concerns here. I'd either remove the else statements or use a case switch. The way I am reading this its either AWS or GCP and only one of these can happen

pkg/clients/aws/route53.go
```diff
@@ -73,107 +72,48 @@ func (c *awsClient) GetDNSName() string {
...
+		zone, err := c.client.GetHostedZone(&route53.GetHostedZoneInput{Id: &fedrampHostedZoneID})
 		if err != nil {
-			reqLogger.Error(err, result.GoString(), "fqdn", fqdn)
+			reqLogger.Error(err, err.Error())
 			return "", err
 		}
-
-		return fqdn, nil
+		input.HostedZoneId = zone.HostedZone.Id
+	} else {
```
rafael-azevedo, line 107: instead of the else here you can set input.HostedZoneId = &dnsZone and overwrite it if it is fedramp

rafael-azevedo (changes requested):

controllers/certificaterequest/issue_certificate.go
```diff
@@ -210,3 +218,22 @@ func (r *CertificateRequestReconciler) IssueCertificate(reqLogger logr.Logger, c
...
+	if err != nil {
+		return "", err
+	}
+
+	if len(dnsZones.Items) != 1 {
+		return "", fmt.Errorf("%d dnsZone objects in a specific namespace found, expected 1 dnsZone", len(dnsZones.Items))
+	}
+
+	if *dnsZones.Items[0].Status.AWS.ZoneID != "" {
+		return filepath.Base(*dnsZones.Items[0].Status.AWS.ZoneID), nil //the format of that ID is (/hostedzone/<something>) so, we don't care about >/. Here, we only want to grab the stuff after the last /.
```
rafael-azevedo, line 234: The comments above were not yet resolved I'd either remove the else statements or use a case switch. The way I am reading this its either AWS or GCP and only one of these can happen

controllers/certificaterequest/issue_certificate.go
```diff
@@ -210,3 +218,22 @@ func (r *CertificateRequestReconciler) IssueCertificate(reqLogger logr.Logger, c
...
+		return "", err
+	}
+
+	if len(dnsZones.Items) != 1 {
+		return "", fmt.Errorf("%d dnsZone objects in a specific namespace found, expected 1 dnsZone", len(dnsZones.Items))
+	}
+
+	if *dnsZones.Items[0].Status.AWS.ZoneID != "" {
+		return filepath.Base(*dnsZones.Items[0].Status.AWS.ZoneID), nil //the format of that ID is (/hostedzone/<something>) so, we don't care about >/. Here, we only want to grab the stuff after the last
+	} else if *dnsZones.Items[0].Status.GCP.ZoneName != "" {
```
fahlmant, line 235:
```suggestion
	}
	if *dnsZones.Items[0].Status.GCP.ZoneName != "" {
```

fahlmant (approved):
/lgtm

controllers/certificaterequest/issue_certificate.go
```diff
@@ -210,3 +218,23 @@ func (r *CertificateRequestReconciler) IssueCertificate(reqLogger logr.Logger, c
...
+	dnsZones := hivev1.DNSZoneList{}
+	err = r.Client.List(context.TODO(), &dnsZones, &client.ListOptions{Namespace: namespace})
+	if err != nil {
+		return "", err
+	}
+
+	if len(dnsZones.Items) != 1 {
+		return "", fmt.Errorf("%d dnsZone objects in a specific namespace found, expected 1 dnsZone", len(dnsZones.Items))
+	}
+
```
iamkirkbater, line 232:
```
dnsZoneStatus := *dnsZones.Items[0].Status

if dnsZoneStatus == nil {
  return "", fmt.Error("Unexpected error, status is not populated")
}

if dnsZoneStatus.AWS != nil {
   // its aws, do something
   return ...
 }
 
 if dnsZoneStatus.GCP != nil {
   // its gcp, do something else
   return ...
 }
 
 return "", fmt.Errorf("Unexpected error: not aws or gcp don't know what to do here")
 ```

PR #273, opened 2024-06-26
remove achvatal from owners

yithian:
:saluting_face:


mjlshen:
/lgtm cancel
@yithian sorry I just noticed, I think you should remove yourself from `OWNERS` instead of `OWNERS_ALIASES` to free yourself :D

PR #274, opened 2024-06-26
updates dns zone logic for DNS challenge to account for FedRAMP

tonytheleg:
For https://issues.redhat.com/browse/OSD-24260
* FedRAMP does not use `DNSZone` CR's nor managed DNS from Hive, the FindZoneID call fails in FedRAMP due to that
* Updates logic to ensure FedRAMP is accounted for

### DIFF
--- a/controllers/certificaterequest/issue_certificate.go
+++ b/controllers/certificaterequest/issue_certificate.go
@@ -114,14 +114,23 @@ func (r *CertificateRequestReconciler) IssueCertificate(reqLogger logr.Logger, c
 		if keyAuthErr != nil {
 			return fmt.Errorf("Could not get authorization key for dns challenge")
 		}
-		dnsZone, err := r.FindZoneIDForChallenge(cr.Namespace)
-		if err != nil {
-			return err
-		}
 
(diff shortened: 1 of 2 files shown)


tnierman:
I think you could potentially achieve the same behavior by wrapping an `else` around [this line](https://github.com/openshift/certman-operator/blob/master/pkg/clients/aws/route53.go#L108), correct?

tonytheleg:
> I think you could potentially achieve the same behavior by wrapping an `else` around [this line](https://github.com/openshift/certman-operator/blob/master/pkg/clients/aws/route53.go#L108), correct?

For the sake of not ignoring this comment in PR since we talked in slack haha

we're failing before we get to the `AWSDnsChallenge` because of `FindZoneIDForChallenge` call not being able to get the hosted zone. I added the fedramp check to make sure we don't pass a value and just let the existing code in `AWSDnsChallenge` get it

The reason for the `else` update is because if I'm reading it right, fedramp or not, that value will get overwritten by that line
```go
	if fedramp {
		zone, err := c.client.GetHostedZone(&route53.GetHostedZoneInput{Id: &fedrampHostedZoneID})
		if err != nil {
			return "", err
		}
		input.HostedZoneId = zone.HostedZone.Id
	}
	input.HostedZoneId = &dnsZone  // <--- this always exectues which would overwrite the value
```

PR #278, opened 2024-07-05
Refactor certman-operator HostedZone Changes

anispate:
Refactor certman-operator HostedZone Changes where we I refactor the FedRAMP logic out of the aws/route53.go file and into the r.FindZoneIDForChallenge() function  as part of [OSD-24304](https://issues.redhat.com//browse/OSD-24304)

### DIFF
--- a/controllers/certificaterequest/issue_certificate.go
+++ b/controllers/certificaterequest/issue_certificate.go
@@ -30,6 +30,7 @@ import (
 	"strings"
 
 	"github.com/go-logr/logr"
+	cClient "github.com/openshift/certman-operator/pkg/clients"
 	"github.com/prometheus/client_golang/prometheus"
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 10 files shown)


fahlmant:
@anispate Can you provide a summary in your PR description of what you specifically changed?

fips.go
```diff
@@ -1,3 +1,4 @@
+//go:build fips_enabled
 // +build fips_enabled
```
fahlmant, line 2: Are both of these needed?

controllers/certificaterequest/issue_certificate.go
```diff
@@ -228,7 +222,14 @@ func (r *CertificateRequestReconciler) IssueCertificate(reqLogger logr.Logger, c
 	return nil
 }
 
-func (r *CertificateRequestReconciler) FindZoneIDForChallenge(namespace string) (string, error) {
+func (r *CertificateRequestReconciler) FindZoneIDForChallenge(namespace string, dnsClient cClient.Client) (string, error) {
+	if fedramp {
+		fedrampZoneid, err := dnsClient.GetFedrampHostedZoneID(fedrampHostedZoneID)
```
fahlmant, line 227: where does the argument here come from?

fahlmant (changes requested):
Please remove the certman-operator binary

controllers/certificaterequest/issue_certificate.go
anispate: cClient We imported the cClient on the file and the fedramp logic, we took from the route53.go file

pkg/clients/aws/route53.go
```diff
@@ -72,6 +72,14 @@ func (c *awsClient) GetDNSName() string {
...
 }
 
+func (c *awsClient) GetFedrampHostedZoneID(fedrampHostedZoneID string) (string, error) {
+	zone, err := c.client.GetHostedZone(&route53.GetHostedZoneInput{Id: &fedrampHostedZoneID})
+	if err != nil {
+		return "", err
+	}
+	return *zone.HostedZone.Id, nil
+}
+
```
tnierman, line 82: It looks like this functions purpose is to get the ID of a `hostedZone` from AWS, but we're already providing an argument called `ID`. Are we returning something different than what's provided? If so, can we rename the function or the arguments to make that a bit clearer?

clcollins (approved):
/lgtm

PR #280, opened 2024-07-15
Return an error when calling AssumeRole

anispate:
As part of [OSD-14258](https://issues.redhat.com/browse/OSD-14258) ticket where Certman silently fails when calling AssumeRole for sts clusters so, we made sure that it returns an error when calling AssumeRole

### DIFF
--- a/pkg/clients/aws/route53.go
+++ b/pkg/clients/aws/route53.go
@@ -611,16 +611,29 @@ func getSTSCredentials(reqLogger logr.Logger, client *sts.STS, roleArn string, e
 		if err == nil {
 			break
 		}
-		if i == (assumeRolePollingRetries - 1) {
-			return nil, fmt.Errorf("timed out while assuming role %s", roleArn)
+		// Error was returned
+		if i >= (assumeRolePollingRetries - 1) {
+			// If we're out of retries, return the wrapped error
+			reqLogger.Info("we are out of retries therefore, we are bailing on this function and returning the wrapped error")
+			return nil, fmt.Errorf("failed to assumeRole, no retries left: %w", err)
 		}
-	}
-	if err != nil {
 		// Log AWS error
 		if aerr, ok := err.(awserr.Error); ok {
 			reqLogger.Error(aerr, "New AWS Error while getting STS credentials,\nAWS Error Code: %s,\nAWS Error Message: %s", aerr.Code(), aerr.Message())
+			// Handle specific AWS errors
+			switch aerr.Code() {
+			case sts.ErrCodeExpiredTokenException:
+				// For expired token errors, stop retrying as it indicates an issue with credentials
+				reqLogger.Info("Token expired, non-retryable error")
+				return nil, fmt.Errorf("expired token error: %w", aerr)
+
+			default:
+				// Handle other AWS errors as needed
+				reqLogger.Info("\"%w\", Unhandled AWS error: %s", aerr, aerr.Code())
+			}
 		}
-		return &sts.AssumeRoleOutput{}, err
+		// If we still have retries, log the failure and try again
+		reqLogger.Error(err, "failed to assumeRole (attempt %d out of %d)", i, assumeRolePollingRetries)
 	}
 	return assumeRoleOutput, nil
 }


pkg/clients/aws/route53.go
```diff
@@ -613,7 +613,7 @@ func getSTSCredentials(reqLogger logr.Logger, client *sts.STS, roleArn string, e
 			break
 		}
 		if i == (assumeRolePollingRetries - 1) {
-			return nil, fmt.Errorf("timed out while assuming role %s", roleArn)
+			return nil, fmt.Errorf("timed out assuming the role, error: %v", err)
```
tnierman, line 616: I think we ought to include the following details:
- Certman tried and failed multiple times to assumeRole, and now we're out of retries
- What errors certman received when attempting to assumeRole. We could just return the last error encountered, like you're doing now, or we could just log them as we go

Very rough & untested code that would satisfy these criteria:
```suggestion
                assumeRoleOutput, err = client.AssumeRole(&assumeRoleInput)
		if err == nil {
		        reqLogger.Errorf("failed to assumeRole (attempt %d out of %d): %v", i, assumeRolePollingRetries, err)
			break
		}
		if i == (assumeRolePollingRetries - 1)  {
			return nil, fmt.Errorf("failed to assumeRole, no retries left: %v", err)
		}
```

tnierman, line 616: I think we ought to include the following details:
- Certman tried and failed multiple times to assumeRole, and now we're out of retries
- What errors certman received when attempting to assumeRole. We could just return the last error encountered, like you're doing now, or we could just log them as we go

Very rough & untested code that would satisfy these criteria:
```suggestion
                assumeRoleOutput, err = client.AssumeRole(&assumeRoleInput)
		if err == nil {
		        reqLogger.Errorf("failed to assumeRole (attempt %d out of %d): %v", i, assumeRolePollingRetries, err)
			break
		}
		if i == (assumeRolePollingRetries - 1)  {
			return nil, fmt.Errorf("failed to assumeRole, no retries left: %v", err)
		}
```

pkg/clients/aws/route53.go
```diff
@@ -613,7 +613,14 @@ func getSTSCredentials(reqLogger logr.Logger, client *sts.STS, roleArn string, e
 			break
 		}
 		if i == (assumeRolePollingRetries - 1) {
-			return nil, fmt.Errorf("timed out while assuming role %s", roleArn)
+			assumeRoleOutput, err = client.AssumeRole(&assumeRoleInput)
+			if err == nil {
+				reqLogger.Error(err, "failed to assumeRole (attempt %d out of %d): %v", i, assumeRolePollingRetries, err)
+				break
+			}
```
mjlshen, line 620:
```suggestion
			if err != nil {
				reqLogger.Error(err, "failed to assumeRole (attempt %d out of %d): %v", i, assumeRolePollingRetries, err)
				break
			}
```

pkg/clients/aws/route53.go
```diff
@@ -613,7 +613,14 @@ func getSTSCredentials(reqLogger logr.Logger, client *sts.STS, roleArn string, e
 			break
 		}
 		if i == (assumeRolePollingRetries - 1) {
-			return nil, fmt.Errorf("timed out while assuming role %s", roleArn)
+			assumeRoleOutput, err = client.AssumeRole(&assumeRoleInput)
+			if err != nil {
+				reqLogger.Error(err, "failed to assumeRole (attempt %d out of %d): %v", i, assumeRolePollingRetries, err)
+				break
```
tnierman, line 619:
```suggestion
				continue
```
Sorry, I think this was a mistake from a suggestion I made prior

pkg/clients/aws/route53.go
```diff
@@ -613,7 +613,14 @@ func getSTSCredentials(reqLogger logr.Logger, client *sts.STS, roleArn string, e
...
+			}
+			if i == (assumeRolePollingRetries - 1) {
```
tnierman, line 621: Correct me if I'm wrong, but it looks like this is the same condition on line 615 - so another error in my suggestion from before. We'll need to rethink this whole stanza slightly, I think

pkg/clients/aws/route53.go
```diff
@@ -611,10 +611,15 @@ func getSTSCredentials(reqLogger logr.Logger, client *sts.STS, roleArn string, e
...
+		// Error was returned
+		if i >= (assumeRolePollingRetries - 1) {
+			// If we're out of retries, return the wrapped error
+			return nil, fmt.Errorf("failed to assumeRole, no retries left: %w", err)
 		}
+		// If we still have retries, log the failure and try again
+		reqLogger.Error(err, "failed to assumeRole (attempt %d out of %d): %v", i, assumeRolePollingRetries, err)
 	}
+
 	if err != nil {
```
iamkirkbater, line 623: Do we still need this block if we're handling the error within the for loop?

anispate:
Added hold status so, we don't want to push anything on Friday and also, we wanted to test https://github.com/openshift/certman-operator/pull/283 PR as well for the weekend.

pkg/clients/aws/route53.go
```diff
@@ -611,16 +611,17 @@ func getSTSCredentials(reqLogger logr.Logger, client *sts.STS, roleArn string, e
...
-			return nil, fmt.Errorf("timed out while assuming role %s", roleArn)
+		// Error was returned
+		if i >= (assumeRolePollingRetries - 1) {
+			// If we're out of retries, return the wrapped error
+			return nil, fmt.Errorf("failed to assumeRole, no retries left: %w", err)
 		}
-	}
-	if err != nil {
 		// Log AWS error
 		if aerr, ok := err.(awserr.Error); ok {
```
fahlmant, line 620: This isn't going to work anymore, as it depends on the err variable that's not removed. This will need to be folded into the for loop now, probably on line 614. Then you can switch on the specific AWS error and either exit immediately, or do a retry

fahlmant (changes requested):

pkg/clients/aws/route53.go
anispate: The err has been mentioned in L607

Also, we don't really need to cast it as an awserr and remove the whole check mentioned below?

```
			// Log AWS error
			if aerr, ok := err.(awserr.Error); ok {
				reqLogger.Error(aerr, "New AWS Error while getting STS credentials,\nAWS Error Code: %s,\nAWS Error Message: %s", aerr.Code(), aerr.Message())
			}
```

fahlmant: Okay, I think I see now that this is handled within the for loop. It still important to cast it as an AWS error IMO to differentiate it from other errors, and as mentioned, depending on what the specific error is from AWS, you could exit the for loop early because retrying would be pointless.

anispate: I assume we are already casting the errors as aws error.  You mean we shall return out the AWS like below?
```
		// Log AWS error
		if aerr, ok := err.(awserr.Error); ok {
			return nil, fmt.Errorf("\"%w\", New AWS Error while getting STS credentials,\nAWS Error Code: %s,\nAWS Error Message: %s", aerr, aerr.Code(), aerr.Message())
		}
```

OR, as you said depending on specific error from AWS, exit loop early so, only way we can achieve it, is by the case statement shown below?

`		if aerr, ok := err.(awserr.Error); ok {
			reqLogger.Error(aerr, "AWS Error while assuming role - Code: %s, Message: %s", aerr.Code(), aerr.Message())

			// Handle specific AWS errors
			switch aerr.Code() {
			case sts.ErrCodeThrottlingException, sts.ErrCodeLimitExceededException:
				// For throttling or limit exceeded errors, continue retrying
				reqLogger.Warn("Retryable AWS error encountered: %s", aerr.Code())
				continue

			case sts.ErrCodeExpiredTokenException:
				// For expired token errors, stop retrying as it indicates an issue with credentials
				return nil, fmt.Errorf("expired token error: %w", aerr)

			default:
				// Handle other AWS errors as needed
				reqLogger.Error("Unhandled AWS error: %s", aerr.Code())
				return nil, fmt.Errorf("AWS error: %w", aerr)
			}`

iamkirkbater: I'd venture a guess as to the second case.

Just the only change I'd make to the logic therein would be having the default case attempt to retry so we don't always _immediately_ exit out.

fahlmant: Kirk got it :)

pkg/clients/aws/route53.go
```diff
@@ -611,16 +611,29 @@ func getSTSCredentials(reqLogger logr.Logger, client *sts.STS, roleArn string, e
...
+			switch aerr.Code() {
+			case sts.ErrCodeExpiredTokenException:
+				// For expired token errors, stop retrying as it indicates an issue with credentials
+				return nil, fmt.Errorf("expired token error: %w", aerr)
+
+			default:
+				// Handle other AWS errors as needed
+				reqLogger.Info("\"%w\", Unhandled AWS error: %s", aerr, aerr.Code())
+				// return nil, fmt.Errorf("AWS error: %w", aerr)
+				continue
```
iamkirkbater, line 632: Extremely small nit - we can remove the commented out return statement _and_ the continue statement here.

_or_

we could leave the Continue, but I'd like to also log the `(attempt %d out of %d)` that we have below.

By adding the `continue` we'll skip the next log message which might still be valuable. since it's retrying.

pkg/clients/aws/route53.go
```diff
@@ -611,16 +611,29 @@ func getSTSCredentials(reqLogger logr.Logger, client *sts.STS, roleArn string, e
...
-	}
-	if err != nil {
 		// Log AWS error
 		if aerr, ok := err.(awserr.Error); ok {
 			reqLogger.Error(aerr, "New AWS Error while getting STS credentials,\nAWS Error Code: %s,\nAWS Error Message: %s", aerr.Code(), aerr.Message())
+			// Handle specific AWS errors
+			switch aerr.Code() {
+			case sts.ErrCodeExpiredTokenException:
+				// For expired token errors, stop retrying as it indicates an issue with credentials
+				return nil, fmt.Errorf("expired token error: %w", aerr)
```
iamkirkbater, line 626: Small nit - a log message before we return like `reqLogger.Info("Token expired, non-retryable error")` or something would be helpful to explicitly mention that we're exiting in the logs.

pkg/clients/aws/route53.go
```diff
@@ -611,16 +611,29 @@ func getSTSCredentials(reqLogger logr.Logger, client *sts.STS, roleArn string, e
...
+			default:
+				// Handle other AWS errors as needed
+				reqLogger.Info("\"%w\", Unhandled AWS error: %s", aerr, aerr.Code())
+				// return nil, fmt.Errorf("AWS error: %w", aerr)
+				continue
+			}
 		}
-		return &sts.AssumeRoleOutput{}, err
+		// If we still have retries, log the failure and try again
+		reqLogger.Error(err, "failed to assumeRole (attempt %d out of %d): %v", i, assumeRolePollingRetries, err)
```
iamkirkbater, line 636: Two options here:

1) We leave this as an Error log type but we can remove the `: %v` part of the log message since it's already going to spit out the `err` as part of the Error logger
2) We change this to be an `reqLogger.Info()` function since it's just informational that we're retrying, still printing the error to the log message for any potential debugging, but not printing out the full stack trace that `reqLogger.Error` would give us.

pkg/clients/aws/route53.go
```diff
@@ -611,16 +611,29 @@ func getSTSCredentials(reqLogger logr.Logger, client *sts.STS, roleArn string, e
 		if err == nil {
 			break
 		}
-		if i == (assumeRolePollingRetries - 1) {
-			return nil, fmt.Errorf("timed out while assuming role %s", roleArn)
+		// Error was returned
+		if i >= (assumeRolePollingRetries - 1) {
+			// If we're out of retries, return the wrapped error
+			return nil, fmt.Errorf("failed to assumeRole, no retries left: %w", err)
```
iamkirkbater, line 617: Suggestion to help our future selves - log an info message here that we're out of retries and are bailing on this function.

Info because the error is being passed back up which _should_ be handled appropriately by the caller function.

iamkirkbater (changes requested):
Some very small changes that I think will greatly help our future selves.

After this I think we're ready to go :D

PR #292, opened 2024-09-04
Adding metrics for UpdateCertificateRetrievalErrors and UpdateMissingCertificates

anispate:
Figuring out bug as part of [OSD-14653](https://issues.redhat.com/browse/OSD-14653)

### DIFF
--- a/pkg/localmetrics/localmetrics.go
+++ b/pkg/localmetrics/localmetrics.go
@@ -85,6 +85,14 @@ var (
 		Name: "cloudflare_failed_requests_count",
 		Help: "Counter on the number of failed DNS requests",
 	})
+	MetricMissingCertificates = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "certman_missing_certificates_total",
+		Help: "Total number of missing certificates by namespace and name",
+	}, []string{"namespace", "name"})
+	MetricCertificateRetrievalErrors = prometheus.NewCounterVec(prometheus.CounterOpts{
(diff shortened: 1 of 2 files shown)


pkg/localmetrics/localmetrics.go
```diff
@@ -193,3 +202,19 @@ func IncrementLetsEncryptMaintenanceErrorCount() {
 func IncrementDnsErrorCount() {
 	MetricDnsErrorCount.Inc()
 }
+
+// register missingCertificates and certificateRetrievalErrors metrics in the prometheus
+func init() {
+	metrics.Registry.MustRegister(missingCertificates)
+	metrics.Registry.MustRegister(certificateRetrievalErrors)
+}
```
iamkirkbater, line 210: Is there a reason we're initializing these two metrics here? How are the other metrics initialized? Should we try to initialize them the same way as the others?

iamkirkbater (changes requested):

pkg/localmetrics/localmetrics.go
anispate: @iamkirkbater  i wasn't able to find anyway they are initializing these metrics the current way therefore, i initalized these new metrics the new way. However, ill be reverting back the new metrics changes once I find the root cause on why current metrics are not getting all the certs inc hibernating, error and LS clusters.

iamkirkbater:
/hold

Should be safe to unhold on Monday, but I'd like to not deploy this so late on a Friday.

/lgtm
/approve

PR #293, opened 2024-09-05
Fixing SecretExists function that returns incorrect values

anispate:
Fixing bug as part of [OSD-25199](https://issues.redhat.com/browse/OSD-25199) ticket

### DIFF
--- a/controllers/certificaterequest/certificaterequest_controller.go
+++ b/controllers/certificaterequest/certificaterequest_controller.go
@@ -392,13 +392,23 @@ func (r *CertificateRequestReconciler) createCertificateSecret(reqLogger logr.Lo
 func (r *CertificateRequestReconciler) revokeCertificateAndDeleteSecret(reqLogger logr.Logger, cr *certmanv1alpha1.CertificateRequest) error {
 	//todo - actually delete secret when revoking
 
-	if SecretExists(r.Client, cr.Spec.CertificateSecret.Name, cr.Namespace) {
-		err := r.RevokeCertificate(reqLogger, cr)
-		if err != nil {
-			return err //todo - handle error from certificate missing
-		}
+	exists, err := SecretExists(r.Client, cr.Spec.CertificateSecret.Name, cr.Namespace)
+	if err != nil {
+		return fmt.Errorf("error checking if secret exists: %w", err)
 	}
+	if !exists {
+		reqLogger.Info("Secret does not exist")
+	}
+
+	error := r.RevokeCertificate(reqLogger, cr)
+	if error != nil {
+		// TODO: handle error from certificate missing
+		return fmt.Errorf("error revoking certificate: %w", error)
+	}
+
+	reqLogger.Info("Certificate successfully revoked")
 	return nil
+
 }
 
 // relocationBailOut checks to see if there's a cluster relocation in progress
--- a/controllers/certificaterequest/utils.go
+++ b/controllers/certificaterequest/utils.go
@@ -20,17 +20,25 @@ import (
 	"context"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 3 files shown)


controllers/certificaterequest/certificaterequest_controller.go
```diff
@@ -390,14 +390,18 @@ func (r *CertificateRequestReconciler) createCertificateSecret(reqLogger logr.Lo
...
+	}
+	if exists {
 		err := r.RevokeCertificate(reqLogger, cr)
 		if err != nil {
-			return err //todo - handle error from certificate missing
+			// TODO: handle error from certificate missing
+			return fmt.Errorf("error revoking certificate: %w", err)
 		}
 	}
+
```
iamkirkbater, line 404: This is a point of personal preference, but I've found that when you have long if statements with a single return at the end it's better to negate and return for a shorter if statement.

For example, comparing the two code snippets:

```
	if exists {
		err := r.RevokeCertificate(reqLogger, cr)
		if err != nil {
			// TODO: handle error from certificate missing
			return fmt.Errorf("error revoking certificate: %w", err)
		}
	}
	return nil
```

versus:

```
	if !exists {
	    reqLogger.Info("Secret does not exist")
	    return nil
	}
	
	err := r.RevokeCertificate(reqLogger, cr)
	if err != nil {
		// TODO: handle error from certificate missing
		return fmt.Errorf("error revoking certificate: %w", err)
	}

	reqLogger.Info("Certificate successfully revoked")
	return nil
```

It's a few more lines now, but I personally find that returning early ends up making the code more clear and easier to maintain over time.

controllers/certificaterequest/certificaterequest_controller.go
```diff
@@ -390,14 +390,18 @@ func (r *CertificateRequestReconciler) createCertificateSecret(reqLogger logr.Lo
 
 // revokeCertificateAndDeleteSecret revokes certificate if it exists
 func (r *CertificateRequestReconciler) revokeCertificateAndDeleteSecret(reqLogger logr.Logger, cr *certmanv1alpha1.CertificateRequest) error {
-	//todo - actually delete secret when revoking
-
-	if SecretExists(r.Client, cr.Spec.CertificateSecret.Name, cr.Namespace) {
+	exists, err := SecretExists(r.Client, cr.Spec.CertificateSecret.Name, cr.Namespace)
```
iamkirkbater, line 393: Do we want to keep the todo here since we're still not deleting any secret after revoking?

controllers/certificaterequest/utils.go
```diff
@@ -20,17 +20,22 @@ import (
...
-
 	err := kubeClient.Get(context.TODO(), types.NamespacedName{Name: secretName, Namespace: namespace}, s)
-	return (err == nil)
+	if err == nil {
+		return true, nil
+	}
+	if errors.IsNotFound(err) {
+		return false, nil
+	}
+	return false, err
```
iamkirkbater, line 38: What is the point of this if statement? Why are we not returning the error here?

(I see the code difference, but maybe this warrants a comment in the code to explain the _why_ for our future selves)

controllers/certificaterequest/utils_test.go
```diff
@@ -0,0 +1,114 @@
...
+	return nil
+}
+
+func (m *MockClient) RESTMapper() meta.RESTMapper {
+	return nil
+}
+
+func (m *MockClient) SubResource(subResource string) client.SubResourceClient {
+	return nil
+}
```
iamkirkbater, line 62: I believe controller runtime has a built-in mock client that we can utilize instead of building our own here.

iamkirkbater (changes requested):

PR #308, opened 2024-10-21
remove config.hypershift check in osde2e test harness

YiqinZhang:
It’s not necessary for the hypershift check to skip the OSDe2e test.
remove
```
if viper.GetBool(config.Hypershift) {
	Skip("Certman Operator is not supported on HyperShift")
}
```
		

[SDCICD-1326](https://issues.redhat.com//browse/SDCICD-1326)


clcollins:
/hold

We're pausing new PRs for a little bit while we get the codebase in order

tnierman:
@clcollins - we can probably `/unhold` at this point, correct?

clcollins:
Yep, I think we're in a much better place now.

/unhold

clcollins (approved):

PR #322, opened 2024-12-05
Fixing the certman_operator_certificate_valid_duration_days metrics

anispate:
As the metrics (https://grafana.app-sre.devshift.net/explore?schemaVersion=1&panes=%7B%22x7t%22:%7B%22datasource%22:%22P3D303D298AFF784A%22,%22queries%22:%5B%7B%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22P3D303D298AFF784A%22%7D,%22editorMode%22:%22code%22,%22expr%22:%22certman_operator_certs_in_last_day_devshift_org%22,%22instant%22:true,%22legendFormat%22:%22__auto%22,%22range%22:true,%22refId%22:%22A%22%7D%5D,%22range%22:%7B%22from%22:%22now-2d%22,%22to%22:%22now%22%7D%7D%7D&orgId=1) are missing here completely and showing us "No data" therefore, we are ensuring that the metrics will show up without showing up the decom certs.

### DIFF
--- a/pkg/localmetrics/localmetrics.go
+++ b/pkg/localmetrics/localmetrics.go
@@ -181,33 +181,18 @@ func AddCertificateIssuance(action string) {
 
 // UpdateCertValidDuration updates the Prometheus metric for certificate validity duration.
 func UpdateCertValidDuration(kubeClient client.Client, cert *x509.Certificate, now time.Time, clusterName, namespace string) {
-	if kubeClient == nil {
-		// If kubeClient is nil, set the metric value based on the certificate's expiration date
-		if cert != nil {
-			diff := cert.NotAfter.Sub(now)
-			days := math.Max(0, math.Round(diff.Hours()/24))
-			MetricCertValidDuration.With(prometheus.Labels{
-				"cn":      cert.Subject.CommonName,
-				"cluster": clusterName,
-			}).Set(days)
-		} else {
-			MetricCertValidDuration.With(prometheus.Labels{
-				"cn":      clusterName,
-				"cluster": clusterName,
-			}).Set(0)
-		}
-		return
-	}
-
-	// Check if the cluster is decommissioned
-	if isClusterDecommissioned(kubeClient, clusterName, namespace) {
-		logger.Info("Cluster is decommissioned, skipping UpdateCertValidDuration", "clusterName", clusterName)
-		return
-	}
-
-	// Set metric based on certificate expiration date if available
 	var days float64
 	var cn string
+
+	// If kubeClient is available, check for decommissioning
+	if kubeClient != nil {
+		if isClusterDecommissioned(kubeClient, clusterName, namespace) {
+			logger.Info("Cluster is decommissioned, skipping metric update", "clusterName", clusterName)
+			return
+		}
+	}
+
+	// Calculate days from certificate
 	if cert != nil {
 		diff := cert.NotAfter.Sub(now)
 		days = math.Max(0, math.Round(diff.Hours()/24))


pkg/localmetrics/localmetrics.go
```diff
@@ -181,31 +181,21 @@ func AddCertificateIssuance(action string) {
...
-				"cn":      cert.Subject.CommonName,
-				"cluster": clusterName,
-			}).Set(days)
-		} else {
-			MetricCertValidDuration.With(prometheus.Labels{
-				"cn":      clusterName,
-				"cluster": clusterName,
-			}).Set(0)
-		}
+	if kubeClient == nil && cert == nil {
```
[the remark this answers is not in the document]
anispate: https://github.com/openshift/certman-operator/blob/a770ff5b1fce61977a8a77ddfa20f0534142943d/pkg/localmetrics/localmetrics.go#L201-L214 <<< That cert is not nil logic is posted below

pkg/localmetrics/localmetrics.go
```diff
@@ -181,31 +181,21 @@ func AddCertificateIssuance(action string) {
...
-			MetricCertValidDuration.With(prometheus.Labels{
-				"cn":      cert.Subject.CommonName,
-				"cluster": clusterName,
-			}).Set(days)
-		} else {
-			MetricCertValidDuration.With(prometheus.Labels{
-				"cn":      clusterName,
-				"cluster": clusterName,
-			}).Set(0)
-		}
```
tnierman, line 198: Why are we losing a conditional with this change? I'm not fully in the loop for this logic, but it reads to me as if we will no longer be pushing a metric if the kubeClient is `nil`, but the cert is not `nil`?

pkg/localmetrics/localmetrics.go
```diff
@@ -181,31 +181,21 @@ func AddCertificateIssuance(action string) {
...
+	if kubeClient == nil && cert == nil {
+		MetricCertValidDuration.With(prometheus.Labels{
+			"cn":      clusterName,
+			"cluster": clusterName,
+		}).Set(0)
```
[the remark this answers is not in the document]
anispate: No, the function does not consider a nil certificate as expired. It simply sets the valid duration metric to 0 days when the certificate is unavailable, without making any assumptions about the actual expiration status.
 the function does not equate a nil certificate to an expired certificate. It simply sets the metric to 0 days in the absence of a valid certificate, without making any further assumptions about the actual expiration status.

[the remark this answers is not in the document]
anispate: The code does skip updating the metric when the cluster is decommissioned. The UpdateCertValidDuration function includes a check for whether the cluster is decommissioned before proceeding to update the metric

`// If kubeClient is available, check for decommissioning
if kubeClient != nil {
    if isClusterDecommissioned(kubeClient, clusterName, namespace) {
        logger.Info("Cluster is decommissioned, skipping metric update", "clusterName", clusterName)
        return
    }
}`

pkg/localmetrics/localmetrics.go
anispate: Updated the code to ensure the metric is always reported, even when the Kubernetes client is not available but a certificate is present.

pkg/localmetrics/localmetrics.go
```diff
@@ -181,33 +181,27 @@ func AddCertificateIssuance(action string) {
...
+	var cn string
+
+	// If both kubeClient and cert are nil, set to 0
+	if kubeClient == nil && cert == nil {
+		MetricCertValidDuration.With(prometheus.Labels{
+			"cn":      clusterName,
+			"cluster": clusterName,
+		}).Set(0)
 		return
 	}
```
tnierman, line 194: Do we actually need this conditional? If I'm reading this logic correctly (and maybe I'm not), this same case is already covered below.

pkg/localmetrics/localmetrics.go
```diff
@@ -181,33 +181,27 @@ func AddCertificateIssuance(action string) {
...
 
-	// Check if the cluster is decommissioned
-	if isClusterDecommissioned(kubeClient, clusterName, namespace) {
-		logger.Info("Cluster is decommissioned, skipping UpdateCertValidDuration", "clusterName", clusterName)
-		return
+	// If kubeClient is available, check for decommissioning
+	if kubeClient != nil {
+		if isClusterDecommissioned(kubeClient, clusterName, namespace) {
+			logger.Info("Cluster is decommissioned, skipping metric update", "clusterName", clusterName)
+			return
```
tnierman, line 200: If a cluster is getting deleted, don't we need to adjust these metrics somehow? ie - set the gauge to zero to indicate the cert isn't valid anymore and/or delete it somehow?

anispate: Currently, we are skipping the update for metrics of deleted or decommissioned clusters instead of setting them to 0 which was the ask from @iamkirkbater  earlier

iamkirkbater: Yeah, explicitly setting it to 0 would result in alerts being fired, we'd need to just remove the metric with those cluster's labels entirely.

Skipping for now works, since they'll just never be updated (meaning they'll never decrement and not fire an alert) and re-deploying the operator would make them go away, but we could create a followup task to get the metrics to actually entirely be removed.

pkg/localmetrics/localmetrics.go
iamkirkbater: When is a case where the kube client would be nil? Do we know this offhand?

iamkirkbater:
Just a few comments, but nothing that should hold this up or change anything so

/approve

PR #326, opened 2025-01-20
Update the 'certman_operator_certificate_valid_duration_days' every time certificaterequest is reconciled

tnierman:
https://issues.redhat.com/browse/OSD-14653

---

Updates the `certificaterequest` controller logic to always update the `certman_operator_certificate_valid_duration_days` metric for `certificaterequests` that aren't being deleted/revoked.

Additionally, new labels `certificaterequest_name` and `certificaterequest_namespace` to the metric clarifying which CR's certificate is being reported on in prometheus.

### DIFF
--- a/controllers/certificaterequest/certificaterequest_controller.go
+++ b/controllers/certificaterequest/certificaterequest_controller.go
@@ -111,23 +111,27 @@ func (r *CertificateRequestReconciler) Reconcile(ctx context.Context, request re
 	err := r.Client.Get(context.TODO(), request.NamespacedName, cr)
 	if err != nil {
 		if errors.IsNotFound(err) {
-			// Request object not found, could have been deleted after reconcile request.
-			// Update metrics to show it's missing and set CertValidDuration to 0
-			localmetrics.UpdateCertValidDuration(r.Client, nil, time.Now(), request.Namespace, request.Namespace)
+			reqLogger.Info("cannot find certificaterequest, assumed deleted")
 			return reconcile.Result{}, nil
 		}
 		// Error reading the object - requeue the request.
 		reqLogger.Error(err, err.Error())
-		localmetrics.UpdateCertValidDuration(r.Client, nil, time.Now(), cr.Namespace, cr.Namespace)
 		return reconcile.Result{}, err
 	}
 
 	// Handle the presence of a deletion timestamp.
 	if !cr.DeletionTimestamp.IsZero() {
-		// Set CertValidDuration to 0 for certificates being deleted
-		localmetrics.UpdateCertValidDuration(r.Client, nil, time.Now(), cr.Namespace, cr.Namespace)
 		return r.finalizeCertificateRequest(reqLogger, cr)
 	}
+
+	defer func(cr *certmanv1alpha1.CertificateRequest) {
+		err := r.UpdateCertValidDuration(cr)
+		if err != nil {
+			reqLogger.Error(err, "failed to update the certman_operator_certificate_valid_duration_days metric")
+		} else {
+			reqLogger.Info("updated certificate duration metrics")
+		}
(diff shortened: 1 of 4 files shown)


controllers/certificaterequest/certificaterequest_controller.go
```diff
@@ -111,24 +111,26 @@ func (r *CertificateRequestReconciler) Reconcile(ctx context.Context, request re
...
 		// Error reading the object - requeue the request.
 		reqLogger.Error(err, err.Error())
-		localmetrics.UpdateCertValidDuration(r.Client, nil, time.Now(), cr.Namespace, cr.Namespace)
 		return reconcile.Result{}, err
 	}
 
 	// Handle the presence of a deletion timestamp.
 	if !cr.DeletionTimestamp.IsZero() {
-		// Set CertValidDuration to 0 for certificates being deleted
-		localmetrics.UpdateCertValidDuration(r.Client, nil, time.Now(), cr.Namespace, cr.Namespace)
```
clcollins, line 128: Does there not need to be any change to the cert metrics if it's deleted?  Or is this cleaned up?

controllers/certificaterequest/certificaterequest_controller.go
```diff
@@ -111,24 +111,26 @@ func (r *CertificateRequestReconciler) Reconcile(ctx context.Context, request re
 	err := r.Client.Get(context.TODO(), request.NamespacedName, cr)
 	if err != nil {
 		if errors.IsNotFound(err) {
-			// Request object not found, could have been deleted after reconcile request.
-			// Update metrics to show it's missing and set CertValidDuration to 0
-			localmetrics.UpdateCertValidDuration(r.Client, nil, time.Now(), request.Namespace, request.Namespace)
 			return reconcile.Result{}, nil
```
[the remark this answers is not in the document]
tnierman: It's a little outside the original scope of these changes, but it's a small enough change that I've added it here :+1:

controllers/certificaterequest/certificaterequest_controller.go
tnierman: In hindsight, this would open us up to an edge-case where a certificate gets close to it's expiry, triggering an alert, but the cluster gets deleted before certman has a chance to renew it. 

In such a case, the metrics would be "locked" at the close-to-expiry value, and the alert would therefore never clear on it's own.

tnierman: It looks like prometheus provides a [DeletePartialMatch](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#MetricVec.DeletePartialMatch) method that allows us to clear these metrics for a specific certificaterequest without having to retrieve the certificate secret (and, therefore, avoiding the potential for a race-condition with hive as cluster objects get cleaned up)

controllers/certificaterequest/certificaterequest_controller.go
```diff
@@ -111,24 +111,26 @@ func (r *CertificateRequestReconciler) Reconcile(ctx context.Context, request re
...
 	// Handle the presence of a deletion timestamp.
 	if !cr.DeletionTimestamp.IsZero() {
-		// Set CertValidDuration to 0 for certificates being deleted
-		localmetrics.UpdateCertValidDuration(r.Client, nil, time.Now(), cr.Namespace, cr.Namespace)
 		return r.finalizeCertificateRequest(reqLogger, cr)
 	}
 
+	defer func(cr *certmanv1alpha1.CertificateRequest) {
+		err := r.UpdateCertValidDuration(cr)
+		if err != nil {
```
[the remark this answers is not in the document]
clcollins: I overlooked it entirely on the first pass, too, for the same reason :D

PR #351, opened 2025-07-22
SREP-1072: added initial set of unit tests for improving coverage

karanjitsingh01:
This PR focuses on improving the test coverage of  packages inside 
Controllers:
- certificaterequest (from 48.8% to 54.0% )
- clusterdeployment (from 66.1% to 71.4%)
- utils (already at 93.9%)

Files Not covered:
certificaterequest/cloudflare.go contains methods that make direct http.Do calls, which are difficult to mock for testing.
certificate_request/revoke_certificate.go relies on the external leclient package, making it difficult to improve test coverage for certain methods. 

The remaining files are now fairly well covered.


controllers/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -492,3 +496,419 @@ func testObjects() []runtime.Object {
...
+				},
+				Status: hivev1.ClusterDeploymentStatus{
+					APIURL:        "https://api.example.com",
+					WebConsoleURL: "https://console.example.com",
+				},
+			},
+			expectGCP:     true,
+			expectSecrets: true,
+		},
+		{
```
bmeng, line 540: The AWS part should be covered by the existing tests.

controllers/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -492,3 +496,419 @@ func testObjects() []runtime.Object {
...
+					ClusterName: "foo",
+					BaseDomain:  "bar.io",
+					ControlPlaneConfig: hivev1.ControlPlaneConfigSpec{
+						ServingCertificates: hivev1.ControlPlaneServingCertificateSpec{
+							Default: "default-cert",
+						},
+					},
+				},
+			},
+			extraRecordEnv: "extra",
```
bmeng, line 673: From the existing coverage, seems the only missing part is when extraRecordEnv is not empty.

controllers/clusterdeployment/clusterdeployment_controller_test.go
```diff
@@ -492,3 +496,419 @@ func testObjects() []runtime.Object {
...
+				return fake.NewClientBuilder().
+					WithScheme(scheme).
+					WithRuntimeObjects(clusterDeployment). // no CRs
+					Build()
+			},
+			expectedNames: []string{},
+			expectError:   false,
+		},
+		{
+			name: "should_return_error_if_list_fails",
```
bmeng, line 798: The same here, seems we only need the error condition to be added.

controllers/certificaterequest/certificaterequest_controller_test.go
```diff
@@ -441,3 +443,54 @@ func TestRelocationBailOut(t *testing.T) {
 		})
 	}
 }
+
+func TestRevokeCertificateAndDeleteSecret(t *testing.T) {
```
bmeng, line 447: Seems we are testing the negative cases only here

controllers/clusterdeployment/clusterdeployment_controller_test.go
karanjitsingh01: removed extra cases

controllers/clusterdeployment/clusterdeployment_controller_test.go
karanjitsingh01: yes, removed  cases that were already covered.
commit id : f1bff7bb79fda8e026a79414070577f013d23173

controllers/certificaterequest/certificaterequest_controller_test.go
karanjitsingh01: Yes, I was not able to cover r.RevokeCertificate, hence only covering error cases for SecretExists method. Let me try to see if I can cover RevokeCertificate using existing testclients or else I can create my own mock client to cover this path.

karanjitsingh01: The main issue is mocking leclient.NewClient(r.Client) in RevokeCertificate
It calls acme.NewClient(directoryURL)  which makes acmeClient.get call, this is internally making REST call. Can you suggest some way to overcome this dependency, one way I can think is to wrap the implementation around interface and then mock it.  your thoughts ?

PR #352, opened 2025-07-23
SREP-1073: Added E2E Test - Check OwnerReference in CertificateRequest

drakshakhan:
This PR adds an end-to-end test to ensure that the ClusterDeployment is set as the owner of the CertificateRequest in the certman-operator namespace. If the owner reference is missing, it is added automatically and verified.

To run the test case, you need to export two environment variables:
DISABLE_JUNIT_REPORT=false
CLUSTER_NAME=<cluster-name>

Then run the test with the command:

### DIFF
--- a/test/e2e/certman_operator_tests.go
+++ b/test/e2e/certman_operator_tests.go
@@ -6,6 +6,7 @@ package osde2etests
 
 import (
 	"context"
+	"os"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -13,15 +14,23 @@ import (
 	configv1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	"github.com/openshift/osde2e-common/pkg/clients/openshift"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 var _ = Describe("Certman Operator", Ordered, func() {
 	var (
+		logger     = log.Log
 		k8s        *openshift.Client
 		clientset  *kubernetes.Clientset
 		secretName string
+
+		dynamicClient dynamic.Interface
+		clusterName string
 	)
 	const (
 		pollingDuration = 15 * time.Minute
@@ -35,6 +44,15 @@ var _ = Describe("Certman Operator", Ordered, func() {
 		Expect(err).ShouldNot(HaveOccurred(), "Unable to setup k8s client")
 		clientset, err = kubernetes.NewForConfig(k8s.GetConfig())
 		Expect(err).ShouldNot(HaveOccurred(), "Unable to setup Config client")
+
+		Expect(clientset).ShouldNot(BeNil(), "clientset is nil")
+		dynamicClient, err = dynamic.NewForConfig(k8s.GetConfig())
+		Expect(err).ShouldNot(HaveOccurred(), "Unable to create dynamic client")
+		Expect(dynamicClient).ShouldNot(BeNil(), "dynamic client is nil")
+
+		clusterName = os.Getenv("CLUSTER_NAME")
+		Expect(clusterName).ToNot(BeEmpty(), "CLUSTER_NAME environment variable must be set")
+
 	})
 
 	It("certificate secret exists under openshift-config namespace", func(ctx context.Context) {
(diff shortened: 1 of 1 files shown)


drakshakhan:
"/label tide/merge-method-squash"

test/e2e/certman_operator_tests.go
```diff
@@ -35,6 +44,15 @@ var _ = Describe("Certman Operator", Ordered, func() {
...
 		clientset, err = kubernetes.NewForConfig(k8s.GetConfig())
 		Expect(err).ShouldNot(HaveOccurred(), "Unable to setup Config client")
+
+		Expect(clientset).ShouldNot(BeNil(), "clientset is nil")
+		dynamicClient, err = dynamic.NewForConfig(k8s.GetConfig())
+		Expect(err).ShouldNot(HaveOccurred(), "Unable to create dynamic client")
+		Expect(dynamicClient).ShouldNot(BeNil(), "dynamic client is nil")
+
+		clusterName = os.Getenv("CLUSTER_NAME")
+		Expect(clusterName).ToNot(BeEmpty(), "CLUSTER_NAME environment variable must be set")
```
bmeng, line 54: Why this must be set? I did not see it is used anywhere in the test.

test/e2e/certman_operator_tests.go
```diff
@@ -62,4 +80,186 @@ var _ = Describe("Certman Operator", Ordered, func() {
...
+			clusterDeployment.SetFinalizers([]string{})
+			_, err = dynamicClient.Resource(clusterDeploymentGVR).Namespace("certman-operator").Update(ctx, &clusterDeployment, metav1.UpdateOptions{})
+			if err != nil {
+				logger.Error(err, "Failed to remove finalizers from ClusterDeployment", "name", cdName)
+				return false
+			}
+
+			logger.Info("Finalizers removed from ClusterDeployment", "name", cdName)
+
+			updatedCD, err := dynamicClient.Resource(clusterDeploymentGVR).Namespace("certman-operator").Get(ctx, cdName, metav1.GetOptions{})
```
bmeng, line 110: As we discussed in one of the sync meet. Lets try to put the clusterdeployment into a separated namespace rather than `certman-operator`

test/e2e/certman_operator_tests.go
```diff
@@ -62,4 +80,186 @@ var _ = Describe("Certman Operator", Ordered, func() {
...
+			if err != nil {
+				logger.Info("ClusterDeployment is deleted", "name", cdName)
```
bmeng, line 112: Why we consider the Clusterdeployment is deleted if the get client request failed in previous step?

test/e2e/certman_operator_tests.go
```diff
@@ -62,4 +80,186 @@ var _ = Describe("Certman Operator", Ordered, func() {
 			return apiserver.Spec.ServingCerts.NamedCertificates[0].ServingCertificate.Name == secretName
 		}, pollingDuration, 30*time.Second).Should(BeTrue(), "Certificate secret should be applied to apiserver object")
 	})
+
+	It("remove finalizer from the ClusterDeployment", func(ctx context.Context) {
```
bmeng, line 84: It will trigger the finalizer deletion on a certain finalizer when the clusterdeployment is deleted.
I cannot quite understand the test here. It seems we try to set the finalizer to empty in the test itself? How would it related to the operator functionality?

test/e2e/certman_operator_tests.go
```diff
@@ -62,4 +80,186 @@ var _ = Describe("Certman Operator", Ordered, func() {
...
+				}
+			}
+
+			logger.Info("Finalizer not yet re-added to ClusterDeployment", "name", cdName)
+			return false
+
+		}, pollingDuration, 30*time.Second).Should(BeTrue(), "ClusterDeployment should be deleted or have finalizer re-added")
+	})
+
+	It("remove finalizer from the CertificateRequest", func(ctx context.Context) {
```
bmeng, line 130: same here, the remove finalizer to certificaterequest happens only when the certificaterequest being deleted.

The function within the test which is setting finalizer to empty string did not test the operator/controller's function.

test/e2e/certman_operator_tests.go
```diff
@@ -62,4 +80,186 @@ var _ = Describe("Certman Operator", Ordered, func() {
...
+				ownerRef := metav1.OwnerReference{
+					APIVersion:         "hive.openshift.io/v1",
+					BlockOwnerDeletion: &isTrue,
+					Controller:         &isTrue,
+					Kind:               "ClusterDeployment",
+					Name:               clusterName,
+					UID:                certRequest.GetUID(),
+				}
+
+				certRequest.SetOwnerReferences(append(ownerRefs, ownerRef))
```
bmeng, line 242: we should not set the ownerReference within the test like this. We need to check the functionality of the operator/controller.

test/e2e/certman_operator_tests.go
drakshakhan: You're right - that was an oversight. A failed Get indicates an error, not deletion. I've fixed it to properly handle errors and only verify the operator re-adds the finalizer.

test/e2e/certman_operator_tests.go
drakshakhan: I've updated the test to verify the operator automatically ensures the finalizer is present when ClusterDeployment is not being deleted, rather than manually removing it. This now properly tests the controller's reconciliation behavior.

test/e2e/certman_operator_tests.go
drakshakhan: Same fix applied - the test now verifies the operator automatically ensures the finalizer is present when not being deleted, rather than manually removing it. This properly tests the controller's functionality.

PR #355, opened 2025-07-24
SREP-1222: Deleted the AWS secret and checked reconciliation in the e2e test.

lambasanchit:
After deleting the aws secret, the certman-operator reconciled the existing CertificateRequest  as expected. During reconciliation, the operator attempted to access the AWS secret but failed due to its absence, resulting in an error logged: Secret "aws" not found. Despite this error, the operator pod remained running without any restarts, which is the intended behavior. No new CertificateRequest was created because no new certificate issuance was triggered. This confirms that the operator gracefully handles missing secrets during reconciliation without crashing or restarting pods.

Reconcile did happen 

Reconcile failed due to missing secret (error logged)

Pod stayed healthy (no restart)

To run the test case, you need to export two environment variables:
DISABLE_JUNIT_REPORT=false
CLUSTER_NAME=cluster-name

Then run the test with the command:


nephomaniac:
Does the failing prow/validate job indicate a need for make boilerplate-update? Thanks.

lambasanchit:
"/label tide/merge-method-squash"

test/e2e/certman_operator_tests.go
```diff
@@ -62,4 +86,102 @@ var _ = Describe("Certman Operator", Ordered, func() {
 			return apiserver.Spec.ServingCerts.NamedCertificates[0].ServingCertificate.Name == secretName
 		}, pollingDuration, 30*time.Second).Should(BeTrue(), "Certificate secret should be applied to apiserver object")
 	})
+
+	Describe("AWS Secret Deletion Scenario", func() {
+
+		It("should ensure ClusterDeployment exists", func(ctx context.Context) {
+			fmt.Println("Starting check for ClusterDeployment")
```
tnierman, line 93:
```suggestion
```
Do we need to print this? Ginkgo's output should already indicate the unit test we're running, right?

test/e2e/certman_operator_tests.go
```diff
@@ -62,4 +86,102 @@ var _ = Describe("Certman Operator", Ordered, func() {
...
+
+			gvr := schema.GroupVersionResource{
+				Group:    "hive.openshift.io",
+				Version:  "v1",
+				Resource: "clusterdeployments",
```
tnierman, line 98: We could use hive's `v1` api to define this: https://pkg.go.dev/github.com/openshift/hive/apis@v0.0.0-20250725035156-a29a23859060/hive/v1#pkg-variables

test/e2e/certman_operator_tests.go
```diff
@@ -1,40 +1,64 @@
...
+		Expect(corev1.AddToScheme(scheme)).To(Succeed())
+
 		var err error
 		k8s, err = openshift.New(GinkgoLogr)
 		Expect(err).ShouldNot(HaveOccurred(), "Unable to setup k8s client")
+
 		clientset, err = kubernetes.NewForConfig(k8s.GetConfig())
-		Expect(err).ShouldNot(HaveOccurred(), "Unable to setup Config client")
+		Expect(err).ShouldNot(HaveOccurred(), "Unable to setup clientset")
+
```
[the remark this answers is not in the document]
lambasanchit: Fixed

test/e2e/certman_operator_tests.go
lambasanchit: Was added for debug .cleaned up now. Thanks

test/e2e/certman_operator_tests.go
lambasanchit: Thanks for the suggestion! I’ve updated it to use hive’s v1 API for clarity

PR #377, opened 2025-08-11
SREP-1220: Add E2E test for revoking and deleting CertificateRequests with certman label

amitupadh:
This PR adds an end-to-end test to validate that when a CertificateRequest contains the label certificaterequests.certman.managed.openshift.io, the following workflow is performed successfully:

-     Revoke the certificate associated with the CertificateRequest.
-     Delete the CertificateRequest after removing its finalizers.
-     Verify reconciliation creates a new CertificateRequest .
-     Ensure the total number of issued certificates remains unchanged before and after the process.

To run the test case, you need to export the below environment variable:
DISABLE_JUNIT_REPORT=false

To run the test:

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -10,7 +10,6 @@ reviewers:
 approvers:
 - clcollins
 - dustman9000
-- iamkirkbater
 - tkong-redhat
 - rbhilare
 - srep-functional-leads


test/e2e/certman_operator_tests.go
```diff
@@ -284,4 +286,98 @@ var _ = ginkgo.Describe("Certman Operator", ginkgo.Ordered, ginkgo.ContinueOnFai
...
+			if err != nil {
+				log.Log.Error(err, "Failed to remove finalizer")
+				return false
+			}
+			return true
+		}, 1*time.Minute, 5*time.Second).Should(gomega.BeTrue(), "Finalizer should be removed")
+
+		// Step 4: Wait for new CertificateRequest with new UID
+		var newCRName string
+		gomega.Eventually(func() bool {
```
tnierman: Unless we're creating a `dnszone.hive.openshift.io` object as a prerequisite to this test and I've missed it (which could very well be the case!), a new certificaterequest would never get recreated just because one was deleted

test/e2e/certman_operator_tests.go
```diff
@@ -284,4 +286,98 @@ var _ = ginkgo.Describe("Certman Operator", ginkgo.Ordered, ginkgo.ContinueOnFai
...
+	ginkgo.It("Delete a labeled CertificateRequest and ensures it is recreated", func(ctx context.Context) {
+		crGVR := schema.GroupVersionResource{
+			Group:    "certman.managed.openshift.io",
+			Version:  "v1alpha1",
+			Resource: "certificaterequests",
+		}
+
+		log.Log.Info("STEP 1: Fetching existing CertificateRequest with owned=true label")
+		crList, err := dynamicClient.Resource(crGVR).Namespace(namespace).List(ctx, metav1.ListOptions{
+			LabelSelector: "certificaterequests.certman.managed.openshift.io",
```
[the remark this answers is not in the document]
karanjitsingh01: Removed the invalid label selector

test/e2e/certman_operator_tests.go
karanjitsingh01: The ClusterDeployment controller Owns CertificateRequest
when a CR is deleted the controller is re-triggered and syncCertificateRequests recreates the missing CR from the ClusterDeployment spec. Domains come from the CD spec (ControlPlaneConfig.ServingCertificates, Ingress) via getDomainsForCertBundle, not from DNSZone.The test ClusterDeployment is created earlier with BuildCompleteClusterDeployment (certificateBundles,controlPlaneConfig, ingress) so the controller has everything it needs to recreate the CR.

PR #386, opened 2025-08-20
SREP-1532 bumping go version to 1.24 and boilerplate for CVE fixes

tkong-redhat:
bumping go version to 1.24 
bumping boilerplate image to v8.1.0 to support 1.24
registry image upgraded to 4.19.
Those are required for CVE fix.

Tested on the staging cluster. certman-operator started without error. Can get dummy certificate via certificate requests


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/openshift/certman-operator
 
-go 1.23.6
+go 1.24.5
```
ravitri, line 3: As discussed, I think we should stick to 4.19 equivalent deps for now with Go 1.23 and Kube 1.32 and controller-runtime v0.20.0. Other deps we can update in next iteration with 4.20 update

PR #476, opened 2026-05-28
Include full certificate chain in TLS secret

dustman9000:
## Summary

One-line fix: change `pemData[0] + pemData[1]` to `strings.Join(pemData, "")` so the full certificate chain is included in the TLS secret.

## Problem

PR #475 added cross-signed chain selection, but the fullchain assembly was hardcoded to only concatenate the first two certs. When the cross-signed chain has three certs (leaf + YR1 intermediate + cross-signed Root YR), the third cert was dropped. Clients couldn't build a trust path back to ISRG Root X1.

## Fix

```go
// Before:
corev1.TLSCertKey: []byte(pemData[0] + pemData[1])

// After:
corev1.TLSCertKey: []byte(strings.Join(pemData, ""))
```

## Incident

[ITN-2026-00169](https://web-rca.devshift.net/incident/ITN-2026-00169)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed certificate chain assembly to properly handle certificate chains with variable numbers of certificates, improving compatibility across different certificate scenarios.

### DIFF
--- a/controllers/certificaterequest/issue_certificate.go
+++ b/controllers/certificaterequest/issue_certificate.go
@@ -206,7 +206,7 @@ func (r *CertificateRequestReconciler) IssueCertificate(reqLogger logr.Logger, c
 	}
 
 	certificateSecret.Data = map[string][]byte{
-		corev1.TLSCertKey:       []byte(pemData[0] + pemData[1]), // create fullchain
+		corev1.TLSCertKey:       []byte(strings.Join(pemData, "")),
 		corev1.TLSPrivateKeyKey: key,
 	}
 


dustman9000:
Closing: the fullchain assembly isn't the issue. FetchAllCertificates found no cross-signed alternate chain from LE, so fetchPreferredChain fell back to FetchCertificates which returns the default 2-cert Gen Y chain. The fix needs to use the classic ACME profile instead.

dustman9000:
Reopening: validated with a full ACME issuance against LE production using an integration cluster domain.

With the `classic` profile (PR #477) and an RSA key (matching certman-operator), LE returns a 3-cert chain:
```
Cert 0: subject=acme-test-rsa.itn-2026-00169.6rmg.i1.devshift.org issuer=YR1
Cert 1: subject=YR1 issuer=Root YR
Cert 2: subject=Root YR issuer=ISRG Root X1
```

The third cert (Root YR cross-signed by X1) is the trust bridge that makes this chain valid for clients with only X1 in their trust store. But `pemData[0] + pemData[1]` drops it, leaving only leaf + YR1 which is untrusted.

This fix (`strings.Join(pemData, "")`) is required alongside #477.

dustman9000:
Consolidated into new PR with classic profile fix.

